Issue #18, opened 2020-08-03
Pagination issues


dawnerd:
Noticed pagination stops working (with react) once you go past the number of items fetched initially and there's no future offset sent to grab more pages. Not sure if this is by design, but it feels more like an oversight.

I've modified my local copy like so:

```js
transformToMeiliSearchParams: function (params) {
      const searchInput = {
        q: params.query,
        facetsDistribution: params.facets.length ? params.facets : undefined,
        facetFilters: params.facetFilters,
        attributesToHighlight: this.attributesToHighlight,
        limit: this.limitPerRequest,
        offset: params.page * this.limitPerRequest
      }
      return removeUndefinedFromObject(searchInput)
    },
```

And removed the page check/splice in `parseHits`. That splice functionality could probably stay if you detected how many were grabbed at a time and the per page limit and did some calculations on that.

curquiza:
Hello @dawnerd, thanks for your report 🙂

I'm don't really understand the issue, but I might have found one concerning pagination (on React only). I notice there is a missing arrow at the bottom of the page:

<img width="854" alt="Capture d’écran 2020-08-06 à 16 33 44" src="https://user-images.githubusercontent.com/20380692/89544904-f62e8700-d802-11ea-9d83-c729d862da04.png">

Is this your issue? 🙂

dawnerd:
Can confirm new version fixed the other issue but not pagination yet

curquiza:
Hello @dawnerd!

I haven't merged the PR that fixes the pagination yet, see #31.
Can you confirm that the problem I describe is the same problem you noticed? Because I'm not sure to exactly understand the pagination issue you have.

curquiza:
@dawnerd 
If your issue is the same issue I described, please replace:

```js
<Pagination />
```
by
```js
<Pagination showLast={true} />
```

The [react codesandbox](https://codesandbox.io/s/ms-react-is-sh9ud?file=/src/App.js) is now up-to-date.

dawnerd:
Nope, didn't help. Your sandbox is still capped at paginating a max of 50 items - solution was in the code block I pasted above.

Every click on a pagination item in the sandbox also returns the same response.

curquiza:
> Every click on a pagination item in the sandbox also returns the same response.

I'm a little bit confused because I don't have this problem when using the codesandbox... It's complicated to fix something I don't succeed to reproduce 😕

curquiza:
[...]

PR #31, opened 2020-08-06
Pagination issue in React and add React playground

curquiza:
Fix this https://github.com/meilisearch/instant-meilisearch/issues/18#issuecomment-669967719.
The change is not visible because I add the react playground at the same time.
I replace:
```js
 <Pagination />
```
by
```js
 <Pagination showLast={true} />
```

I don't know if it fixes the complete issue yet (waiting for the author's answer).

Plus: Add a React Playground

NB about the React Playground:
- `playgrounds/react/src/instant-meilisearch.js` is a soft link, otherwise react-add does not want to start (it can not import files outside of the `src` folder by default)
- `SKIP_PREFLIGHT_CHECK=true` in the `.env` file allows the app to start even if the version of `jest-babel` is not the same in the React playground and in the main app (at the root of the repo)


playgrounds/react/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "react-playground",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@testing-library/jest-dom": "^4.2.4",
+    "@testing-library/react": "^9.3.2",
+    "@testing-library/user-event": "^7.1.2",
+    "babel-jest": "^26.2.2",
```
bidoubiwa, line 9: As these dependencies are not needed in production they should be added in the `devDependencies`. As such they will not be downloaded in production mode. I realise that this is mostly will not happen here but it is best practice and if someone copies your playground he could think that because they are in `dependencies` they are needed in production mode.

bidoubiwa: Also, they are not needed because they are no tests

playgrounds/react/src/App.js
```diff
@@ -0,0 +1,76 @@
...
+            <Hits hitComponent={Hit} />
+            <Pagination showLast={true}/>
+          </div>
+        </InstantSearch>
+      </div>
+    );
+  }
+}
+
+function Hit(props) {
```
bidoubiwa, line 63: Is placeholder search possible?

bidoubiwa:
I think they might still be a problem with the pagination. When I type `war` and go rapidly from page 6 to 7 and 7 to 6. The pages sometimes needs to refresh completely and then ignores my change. 

I'm not sure this is an important bug but still letting you know

playgrounds/react/src/App.js
curquiza: Not yet, this is this PR #34

playgrounds/react/package.json
curquiza: Ok, I just followed the getting started of React to create a basic react application: https://create-react-app.dev/docs/getting-started/

bidoubiwa: Allright ! As you wish it's a detail :)

playgrounds/react/src/App.js
bidoubiwa: Ooke !

playgrounds/react/package.json
curquiza: I removed them!

curquiza:
What do you mean by "ignores your changes"?

curquiza:
@bidoubiwa do you have the same issue on Algolia's playground? https://codesandbox.io/embed/github/algolia/doc-code-samples/tree/master/React+InstantSearch/getting-started

bidoubiwa:
@curquiza Yes !

curquiza:
Hope this is not annoying for our users...

PR #35, opened 2020-08-10
Add scripts in packages.json for demo tests

curquiza:
What do you think @bidoubiwa? Is the naming correct for the `scripts` in `package.json` file?


bidoubiwa:
LGTM

PR #96, opened 2020-10-15
Update the package according to MeiliSearch v0.15.0

curquiza:


src/index.js
```diff
@@ -21,13 +21,13 @@ export default function instantMeiliSearch(hostUrl, apiKey, options = {}) {
...
-        q: this.placeholderSearch && query === '' ? null : query,
+        q: query,
         facetsDistribution: facets.length ? facets : undefined,
-        facetFilters: facetFilters,
+        facetFilters,
         attributesToHighlight: this.attributesToHighlight,
         attributesToCrop: attributesToSnippet,
         filters,
-        limit,
+        limit: this.placeholderSearch === false && query === '' ? 0 : limit,
```
bidoubiwa, line 30:
```suggestion
        limit: this.placeholderSearch === false && query ? 0 : limit,
```
because `''` is a falsy, you can do `if (query)`. That way you cover all possibilities for the query value, see doc: 
https://developer.mozilla.org/fr/docs/Glossaire/Falsy

You would cover the following query values with `if (query)`
- `false`
- `null`
- `undefined`
- `0`
- `NaN`
- `''`
- `""`

You could also do `if (this.placeholderSearch)` but maybe you prefer the clarity of the code?  Your call.
```suggestion
        limit: (this.placeholderSearch && query) ? 0 : limit,
```

bidoubiwa (changes requested):

src/index.js
curquiza: I don't want to cover more than an empty string, because that's what instantsearch sends when there is no query at all in the input.

curquiza: except if `''` and `""` are different in JS (but I'm pretty sure that's the same), I would need to write 
```js
... && (query === '' || query === "") ...
```

bidoubiwa: No they are not considered different. No problem then you can ignore my suggestion :)

bidoubiwa (approved):

PR #138, opened 2020-11-12
Add attibutesToRetrieve

### LINKED ISSUE #137: Add attributesToRecieve as an param
Hey guys, first of all thanks for the awesome work you provided!

I have a little question: As described [in your Documentation](https://docs.meilisearch.com/guides/advanced_guides/search_parameters.html#attributes-to-retrieve) there is an option to specify which attributes to receive from a query. I tried to configure it, but sadly it isn't considered in the `transformToMeiliSearchParams`-Function

https://github.com/meilisearch/instant-meilisearch/blob/master/src/index.js#L16-L22

Was there a specific reason to not include this? If not, is there a possibility to add this in a future release?

Thanks!

curquiza:
Fix #137

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,6 +18,7 @@ export default function instantMeiliSearch(hostUrl, apiKey, options = {}) {
         facets,
         facetFilters,
         attributesToSnippet,
+        attributesToRetrieve,
         filters,
       } = params
       const searchInput = {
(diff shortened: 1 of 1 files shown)


bidoubiwa (approved):
LGTM

curquiza:
bors merge

Issue #122, opened 2020-11-05
Named export vs Default export


bidoubiwa:
# Named or Default

As of today, `instant-meilisearch` uses default export. This means that in order for a user to use instant-meilisearch in an ES environment he has to import it this way: 
```javascript
import instantMeiliSearch from 'instant-meilisearch'
```
The suggestion this issue considers is changing from this default export to a named export:
```javascript
import { instantMeiliSearch } from 'instant-meilisearch'
```

Should we change our default export to a named export? 
Here are a few arguments, if you have any other feel free to share it here/

## Pro's  of default search

- More common, or at least they used to be more common. There is no exact data on it. Although, I observe a change in this tendency. 
- Can be considered **Easier to use**, as only one export where you can change the name offers no complication to use it. `import whatIwant from 'instant-meilisearch'` instead of `import { instantMeiliSearch } from 'instant-meilisearch'`.
    Which can also be done with named exports `import { instantMeiliSearch as whatIWant } from 'instant-meilisearch'`, which if you use VS code should be auto-completed and thus present no difficulties to do.
- Changing it now would be breaking change (but the earlier, the better?) @bb 
- It seems that Typescript suggest using default exports in the [red flag category](https://www.typescriptlang.org/docs/handbook/modules.html#red-flags) of their documentation. But this is not really clear and based on [this stackOverFlow post](https://stackoverflow.com/questions/33305954/typescript-export-vs-default-export#:~:text=For%20example%2C%20if%20you%20default,in%20all%20the%20other%20files) where a user that prefers named exports mentions the redflag.
- No changes needed in the code as it is already a default export
- [vue-instantsearch](https://github.com/algolia/vue-instantsearch) and [algoliaSearch](https://github.com/algolia/instantsearch.js) uses default export **but not** [react-instantsearch](https://github.com/algolia/react-instantsearch)
    ```
    import InstantSearch from 'vue-instantsearch';
    import algoliasearch from 'algoliasearch/lite';
    import { InstantSearch, SearchBox, Hits } from 'react-instantsearch-dom';
    ```
## Con's of default search

The first 5 arguments comes from this [typescript book](https://github.com/basarat/typescript-book) that has 12k stars on github we learn that:  [default is bad](https://basarat.gitbook.io/typescript/main-1/defaultisbad). I would suggest reading the article as it showcases it better.

[...]

Issue #121, opened 2020-11-05
Converting the project to typescript


bidoubiwa:
## Proposition

As this is a small project, converting it to typescript is a task that should not be that hard.

## Benefits 

If the project is in typescript, we can generate a `typed.d.ts` file containing all the types from this project. This file, in turn, would make this library available for TypeScript user. 

More types mean fewer errors, an easier coding experience, and overall better readability of the code. 

## Cons

Adding typescript brings some complexity in configuration and requires more knowledge to maintain than a vanilla JS project.

Related to #118

bidoubiwa:
Implemented by: #249

PR #139, opened 2020-11-12
Convert project to TypeScript

### LINKED ISSUE #118: Usage with TypeScript
I'd like to use this library with TypeScript, however there do not seem to be official typings available.

Are there any plans to add typings?
If so, which would be the preferred way?
One could:
* convert the project to TypeScript (as e.g. meilisearch-js
already is in TypeScript as well),
* add separate d.ts files here inside this repository or
* add them outside e.g. at DefinitelyTyped.

<details>
<summary>Related issue which might be solved by the above</summary>

Besides the actual types I had another problem: even with the usual no-types-available approach (adding a separate .d.ts file with `declare module "@meilisearch/instant-meilisearch";`), I could not immediately use the lib with TypeScript and Webpack. Typescript claimed the Module is no function. I worked around it by modifying the source directly and exporting the function (instead of default export).

</details>

bb:
This is a rough conversion of the project to TypeScript. With rough I mean some usage of `any` here and there, some ad-hoc type definitions and many eslint rules disabled. This fixes #118 and #121. It also sneaks in the fix for Contributing Docs (same as meilisearch/meilisearch-js#687).

Everything should work at least as good as before. It still uses the default export (#122) because the way it is written now, it can be used in another TypeScript project.


bb:
The failing integration-test is because https://github.com/meilisearch/meilisearch-js/pull/688 is not yet released.

If you check out meilisearch-js locally and use it by running `yarn link` in meilisearch-js and `yarn link meilisearch` in instant-meilisearch, it will work.

bb:
bors try

curquiza:
bors try

curquiza:
Hello @bb!
Thanks a lot for this work! This is a huge start to move this project to typescript!!
Because it might need some improvements, we created a `typescript-dev` branch and we will merge your work into this branch as a first step 🙂
@bidoubiwa and I will review your PR and will continue your work on the `typescript-dev` branch! When all will be ready, we will merge `typescript-branch` into `master`.
Thanks again for everything!!! 🚀

curquiza (commented):
Thanks for this huge work! We are waiting for the @bidoubiwa's review.

A small change maybe if you have time @bb: could you remove the changes in the CONTRIBUTING.md? It's not related to the typescript work.

src/index.ts
```diff
@@ -1,18 +1,101 @@
...
+      hits: any
+      index: string
+      hitsPerPage: number
+      facets: object | undefined
+      exhaustiveFacetsCount?: boolean | undefined
+      processingTimeMs: number
+      exhaustiveNbHits: boolean
+    }>
+  }>
+}
```
bidoubiwa, line 82: Should these types not be exported as well?

src/format.ts
```diff
@@ -1,6 +1,11 @@
-import { isString } from './utils.js'
+import { Params } from '.'
+import { isString } from './utils'
 
-function replaceHighlightTags(value, highlightPreTag, highlightPostTag) {
+function replaceHighlightTags(
+  value: string,
+  highlightPreTag?: string,
+  highlightPostTag?: string
+) {
```
bidoubiwa, line 8: The return type is missing or is it infered based on the returned function?

src/format.ts
```diff
@@ -15,30 +20,30 @@ function replaceHighlightTags(value, highlightPreTag, highlightPostTag) {
   return newHighlightValue.toString()
 }
 
-function createHighlighResult({
+function createHighlighResult<T extends Record<string, any>>({
   formattedHit,
   highlightPreTag,
   highlightPostTag,
-}) {
+}: { formattedHit: T } & Params) {
```
bidoubiwa, line 27: Why is it both `{ formattedHit: T }` and `Params` as return types?

bidoubiwa (commented):
Thanks a lot for this work! I see you took the configuration of meilisearch-js. Can I ask you if you think some things should be improved in the typescript configuration? In order to know on what to work. 

I asked you some questions I hope they are relevant

bb:
> A small change maybe if you have time @bb: could you remove the changes in the CONTRIBUTING.md? It's not related to the typescript work.

Good point. Sorry for mixing this stuff. I now committed the contributing file from current master on top, so all should be fine again.

src/format.ts
bb: Because formattedHit is not defined in Params.

If you remove it, line 24 will be marked as error:
> Property 'formattedHit' does not exist on type '{ page?: number | undefined; hitsPerPage?: number | undefined; highlightPreTag?: string | undefined; highlightPostTag?: string | undefined; snippetEllipsisText?: string | undefined; attributesToSnippet?: string[] | undefined; query: string; facets: string[]; } & SearchParams<...>'.ts(2339)

However, it would probably be a good idea to refactor this into a separate named type. One of the rough edges I mentioned in the initial description 😉

src/format.ts
bb: Yes, it's inferred. The return statement explicitly states `.toString()`.

However, there are other things which might not be perfect yet: Is the input param really always string as specified? Currently it looks like, but why would it need to check isString() in line 15 then?

src/index.ts
bb: I didn't need them, but they might be useful to someone. Should I add the exports or should we wait until they're requested?

bb:
> Thanks a lot for this work! I see you took the configuration of meilisearch-js. Can I ask you if you think some things should be improved in the typescript configuration? In order to know on what to work.

I thought it would be a good idea to have both projects not diverge too much, so I copied some configs from there. I think the config itself is fine.

Things to improve in general:
* there's one Todo comment left (index.ts line 14)
* there are quite a few `any` type annotations, esp. in `InstantMeiliSearchInstance`
  * an especially nasty one is `(result[key] as any)` in format.ts. I didn't find a good solution within a few minutes of googling, so I left it as is.
* there are a few `foo as Bar` casts, which are probably also a code smell. 
  * e.g. the `Object.keys(object) as K[])` in utils.ts line 9
* It would probably be nice to enable generic Parametrization of the Index's structure (which currently uses `    meiliSearchResponse: SearchResponse<any, any>`)

As you can see, it's all not critical to fix but also not a great clean state yet.

Does this overview help as a starter?

src/index.ts
curquiza: No need to export them for this first PR 🙂 Let's keep it like this for now.

curquiza (approved):
I also realize we have to find a way to activate the "auto-reload/rebuild"  for the playgrounds because of this typescript integration, otherwise, it will be tedious to develop with them.

Thanks for your work @bb, there is only a small git conflict (I made the `typescript-dev` breach up-to-date with `master`).
Afterward, I'm going to merge your PR. @bidoubiwa will continue your work. She's on vacations right now and has emergency tasks on her todo list before doing this one, but be sure she's looking forward to working on it 😉
And thanks for taking the time to answer our questions and to suggest improvements! 🚀

bidoubiwa (approved):
Thanks for the amazing work

bb:
I just fixed the merge conflict in package.json and modified the new vanilla-js taks to include the build step directly in the Github UI, but it looks like this was not enough. Will check locally now.

bb:
I resorted to an entry in the ignore file to make the linter pass with the TypeScript base settings and the new vanilla-js parts. Not perfect, but better than failing checks.

curquiza:
Merging then!! Thanks a lot @bb for all your involvement!! 🎉

PR #216, opened 2021-01-16
Add error information in console

bidoubiwa:
Best I could do given that the problem comes from algolia itself: 

Screenshot 2021-01-16 at 23 28 02

First error shown is our error, second is the unhandled error of algolia. Maybe a more deep dive search could help us find a solution where algolia's unhandled error would not appear. But meanwhile, the user is aware of its errors.

fixes https://github.com/meilisearch/meilisearch-vue/issues/7
fixes #216

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -92,23 +92,28 @@ export default function instantMeiliSearch(hostUrl, apiKey, options = {}) {
     },
 
     search: async function (requests) {
-      // Params got from InstantSearch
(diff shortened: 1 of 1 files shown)


curquiza (approved):
Awesome! Thanks @bidoubiwa!

CaroFG (approved):

bidoubiwa:
bors merge

PR #239, opened 2021-01-27
Fix node-forge vulnerability in playgrounds/html

curquiza:
And update dependencies in `playgrounds/html/yarn.lock`

### DIFF
--- a/playgrounds/html/package.json
+++ b/playgrounds/html/package.json
@@ -21,5 +21,8 @@
   "devDependencies": {
     "@babel/core": "^7.12.10",
     "parcel-bundler": "^1.12.4"
+  },
+  "resolutions": {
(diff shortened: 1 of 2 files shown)


bidoubiwa:
I do not see `node-forge` in master `yarn.lock`

bidoubiwa:
An issue was opened in the [parcel repo](https://github.com/parcel-bundler/parcel/issues/5250)

> You do realize that this isn't an actual issue right? You have to use `nodeforge.util.setPath` in order to trigger the vulnerability, which parcel does not. These npm warnings just create work for maintainers in most cases due to people complaining in issues like this one.

So we might want to merge just to remove the vulnerability in github interface.

bidoubiwa (approved):

curquiza:
bors merge

PR #247, opened 2021-02-02
Fix docs links

react-learner:
New documentation file structure means paths have changed, so I updated the links.


curquiza (approved):
Thanks @react-learner!

bors merge

PR #256, opened 2021-02-09
Add angular playground

bidoubiwa:


.github/workflows/test.yml
```diff
@@ -26,6 +26,9 @@ jobs:
         run: yarn build
       - name: Run env tests
         run: yarn test:env
+      - name: Run env tests
+        run: yarn test:env
+
```
curquiza, line 31:
```suggestion
```

package.json
```diff
@@ -6,21 +6,25 @@
   "scripts": {
     "cleanup": "shx rm -rf dist/",
     "test": "yarn build && jest",
-    "test:all": "yarn test && yarn test:env && yarn test:playgrounds",
+    "test:all": "yarn test && yarn test:env && yarn test:playgrounds && yarn test:e2e",
```
curquiza, line 9: Do you add the `yarn test:e2e` of angular because you plan to add other e2e tests? Otherwise this is only a test that should be in `test:playgrounds` instead of `test:all`, right?

playgrounds/angular/README.md
```diff
@@ -0,0 +1,25 @@
...
+```
+
+## Running end-to-end tests
+
+```
+yarn e2e
+```
+
+Run `yarn e2e` to execute the end-to-end tests via [Protractor](http://www.protractortest.org/).
+
```
curquiza, line 25:
```suggestion
```

package.json
bidoubiwa: Playground tests are basically "do the playground build". Where as the `e2e` tests really tests `instant-meilisearch` the same way we test all meilisearch SDK's. 

Also, yes, i would like in the long run to have e2e tests for instantsearch envs: react, vue, js.

curquiza (approved):

bidoubiwa:
bors merge

PR #322, opened 2021-03-11
Add objectID when primaryKey is provided

bidoubiwa:
## Problem

With the new version of `instantSearch`, `objectID` becomes mandatory in the hits information to avoid errors and warnings in the console.

For example, in `React` when using instant-meilisearch the following error is raised: 
```
react_devtools_backend.js:2430 Warning: Failed prop type: The prop `hits[0].objectID` is marked as required in `Hits`, but its value is `undefined`.

react_devtools_backend.js:2430 Warning: Each child in a list should have a unique "key" prop.
```

Because instantSearch adds the `objectID` as the key of every hit component, we both receive an error for its absence and an error because our components are missing a key prop as [react expects a key on every component](https://reactjs.org/docs/lists-and-keys.html#keys).

## Solution

First I tried to fetch the primaryKey from MeiliSearch but the public key does not give permission to fetch the primaryKey.

So the solution was to add the `primaryKey` option to instant-meilisearch options. If it is not present, the `objectID` field will not be added in the returned hits. If a wrong primaryKey is given its value will be undefined and the error above will be raised the same way.

### Usage 

```javascript
  'https://demos.meilisearch.com',
  'dc3fedaf922de8937fdea01f0a7d59557f1fd31832cb8440ce94231cfdde7f25',
  {
    paginationTotalHits: 60,
    primaryKey: 'id',
  }
)
```

Not breaking!

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ const searchClient = instantMeiliSearch(
   {
     paginationTotalHits: 30, // default: 200.
     placeholderSearch: false // default: true.
+    primaryKey: 'id' // default: undefined
   }
 );
 ```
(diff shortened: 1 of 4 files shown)


mdubus (approved):

curquiza (commented):
Is it the best way to handle this error? I mean, if I understand well, the user must set "primaryKey" when setting instantMeiliSearch, otherwise, the error will still be present. It implies user manipulation to remove the error, which is not really convenient.

What about always sending to InstantSearch an `objectID` field containing nothing (or whatever) to remove the error without any user manipulation? (of course the case of the user having an "objectID" field in his/her document should be handled)
I'm not sure this is valuable for the user to have two fields with the same value, even if, I agree, `objectID` seems to be an identifier.

EDIT: Hum I've just realized my suggestion is bad because each `objectID` value has to be unique, right?

bidoubiwa:
> Hum I've just realized my suggestion is bad because each objectID value has to be unique, right?

Yes we thought of: 
- Creating a random number but random means it could be the same. Unless we use a [specific library to generate a complex number](https://www.npmjs.com/package/uuid). Could add processing time, but mostly adds a library for the user to download when arriving on a website using instant-meilisearch.
- Hashing the hit to have a unique value based on the content of the hit. But this adds processing time between each search.

As this error could be potentially ignored by our users, as the search experience still works, only the ones interested in removing the error would use this option. 
It is optional and could be ignored by users who do not need it.

curquiza:
Let's wait for the possibility of fetching the `primaryKey` with the public key then...
Thanks for explaining all the ideas you thought! 🙂

curquiza (approved):

curquiza:
@mdubus thanks for your time on the reviews, this is really appreciated ❤️

bidoubiwa:
bors merge

PR #323, opened 2021-03-11
Fix processing time that was always 0 in Stats widget

bidoubiwa:
Processing time provided to instantSearch was badly formatted, as instantSearch expects `processingTimeMS` and we provided `processingTimeMs` as per MeiliSearch naming in its response.


curquiza (approved):

bidoubiwa:
bors merge

PR #344, opened 2021-03-25
Fix nested object being transfom into [object, object] in highlight

bidoubiwa:
The body of _highlightedResult used to look like this:

Capture d’écran 2021-03-25 à 14 42 51

The field should be transformed to String using `JSON.stringify` to keep all information. 

Screenshot 2021-03-25 at 15 20 33
It looks like this with this fix.


mdubus (approved):

curquiza (approved):

bidoubiwa:
bors merge

Issue #389, opened 2021-05-03
Needs updated: Compatibility with algoliaSearch parameters


bidoubiwa:
# Compatibility

All bulletpoints that have been crossed are UI related and do not need to be tested.

Not compatible features list at the end of the issue.
Not compatible parameters start with a ❌.

## search

- query: The text to search in the index
❗️ Not tested because completely ignored by instantSearch

- similarQuery: Overrides the query parameter and performs a more generic search that can be used to find “similar” results.
❗️ Not tested because usefulness is questionable

## attributes

- ✅ attributesToRetrieve: Gives control over which attributes to retrieve and which not to retrieve.
  - In MeiliSearch `attributesToRetrieve` does not impact `attributeToHighlight`

- ❌ restrictSearchableAttributes: Restricts a given query to look in only a subset of your searchable attributes.
    This setting overrides searchableAttributes for specific searches, without changing the priority order of the searchableAttributes.

## ranking

- ❌ relevancyStrictness: Controls the relevancy threshold below which less relevant results aren’t included in the results.

## filtering
- ✅ filters: Filter the query with numeric, facet and/or tag filters. #393
    - ✅ 'attribute = value AND | OR | NOT attribute = value' // Different usage than [instantSearch Docs](https://www.algolia.com/doc/api-reference/api-parameters/filters/#examples)
    - ✅ 'numeric_attribute = | != | > | >= | < | <= numeric_value'
     -   ❌ 'attribute:lower_value TO higher_value'
     -   ❌ 'facetName:facetValue'
     -   ❌ '_tags:value'
     -   ✅ 'attribute = value'  // Different usage than [instantSearch Docs](https://www.algolia.com/doc/api-reference/api-parameters/filters/#examples)

- ✅ facetFilters: Filter hits by facet value. #400

-   ❌ optionalFilters: Create filters for ranking purposes, where records that match the filter are ranked higher, or lower in the case of a negative optional filter.

- ✅ numericFilters: Filter on numeric attributes. #393

-  ❌ tagFilters: Filter hits by tags.

-  ❌ sumOrFiltersScores: Determines how to calculate the total score for filtering.

## faceting
- ✅ facets: Facets to retrieve. #400

- ❌  maxValuesPerFacet: Maximum number of facet values to return for each facet during a regular search.
    If you want to change the number of retrieved facet hits during a search for facet values, see maxFacetHits.

this is possible using `limit` in `refinement-list`
Example:
[...]

PR #387, opened 2021-05-03
Add pagination tests + fix pagination

bidoubiwa:

PR #408, opened 2021-05-24
Add snippet/highlight/pagination integration tests

bidoubiwa:
see #389 

Unit tests have also been added see: 
#387

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -35,7 +35,11 @@ export type IMResponse = {
   nbPages?: number
 }
 
-export type SearchResponse = IStypes.SearchResponse & IMResponse
+export type SearchResponse = Omit<IStypes.SearchResponse, 'hits'> &
+  IStypes.SearchResponse &
+  IMResponse & {
+    hits: ISHits[]
+  }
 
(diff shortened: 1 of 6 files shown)


src/types/types.ts
```diff
@@ -35,7 +35,11 @@ export type IMResponse = {
   nbPages?: number
 }
 
-export type SearchResponse = IStypes.SearchResponse & IMResponse
+export type SearchResponse = Omit<IStypes.SearchResponse, 'hits'> &
```
curquiza, line 38: Just for my knowledge, what does the `&` mean? (sorry not to have asked before, I'm sure it's obvious for you)

curquiza (approved):
Thanks a lot @bidoubiwa for these tests, you save us!

src/types/types.ts
bidoubiwa: It is an intersection between the types ([see doc for great examples](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces))

bidoubiwa:
bors merge

PR #458, opened 2021-07-28
Add cache

bidoubiwa:
Will cache all fields present in provided filter (ex: when clicking on the checkbox of a facet). If the returned response from MeiliSearch's facetsDistribution does not contain one of the fields from the filter it will be added with a value of 0. 

This avoid the following problem:

no_fix

Solution: 

fix

see meilisearch/meilisearch-js/issues/970


alallema (approved):
LGTM! 🎉 
But if it's for the v0.21 why `facetFilters` still exist ?

bidoubiwa:
It should be removed! absolutely :) I will do that

alallema (commented):
There are also some `filters` instead of `filter`

bidoubiwa:
Yes this is normal, it is because instantsearch expects facetFilters and filters
But there were places where I had to change them so thanks :)

PR #504, opened 2021-08-25
Fix umd build having no node interop

bidoubiwa:
The bug was caused by changing the rollup typescript from the non-official [`rollup-plugin-typescript2`](https://www.npmjs.com/package/rollup-plugin-typescript2)  to the official one [@rollup/rollup-typescript](https://github.com/rollup/plugins/tree/master/packages/typescript/#readme).

Multiple libraries suggesting multiple bundles outputs (umd, cjs, ...) are using rollup-plugin-typescript2. Since some of the issues why we initially used this one were resolved, I made the change to go back to the official one. 
Unfortunately do a lack of tests on the builds themselves I did not realise a new issue was introduced. 

By rollbacking to the initial plugin we used, the problem seems to be fixed. Tests are added to ensure this.


alallema (approved):
LGTM!

bidoubiwa:
bors merge

PR #592, opened 2021-11-08
Update playground meilisearch hosts to new host

bidoubiwa:


curquiza (approved):
I trust you this is the public key and not the master one ;)

Also don't forget the skip-changelog label

PR #630, opened 2022-01-06
Add types tests

bidoubiwa:


package.json
```diff
@@ -54,7 +55,7 @@
     "url": "https://github.com/meilisearch/instant-meilisearch.git"
   },
   "dependencies": {
-    "meilisearch": "^0.23.0"
+    "meilisearch": "0.23.0"
```
alallema, line 58: It's normal that is not the `v0.24.0`?

PR #638, opened 2022-01-12
Fix assertions on cypress tests after meili v0.25 update

brunoocasali:
After the v0.24 -> v0.25 upgrade we could see a difference in how objects are ordered inside the list (the same behavior appeared on rails integration tests https://github.com/meilisearch/meilisearch-rails/pull/99/files#r783346153).

Beyond the fix, this test tries to rely on just checking the presence of one object to ensure the behavior is kept, I tried locally to get the first result and the first result after the action, but I failed miserably 😭.

This commit solves the problem until the first object would be affected by a future upgrade on Meili.

Hope this could help to release the v0.25 version for instant-meilisearch 🥇

### DIFF
--- a/cypress/integration/search-ui.spec.js
+++ b/cypress/integration/search-ui.spec.js
@@ -57,14 +57,15 @@ describe(`${playground} playground test`, () => {
   })
 
   it('click on facets', () => {
+    cy.get(HIT_ITEM_CLASS).eq(0).contains('Counter-Strike')
+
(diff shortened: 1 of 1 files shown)


curquiza:
bors try

brunoocasali:
bors try

curquiza:
bors try

mdubus (approved):
LGTM 🔥

PR #649, opened 2022-01-18
Update credentials

curquiza:
Update with new default search key
I've updated the instance to use MeiliSearch v0.25.2, and I used a dump coming from v0.24.0, so the API keys were reset.

To merge as soon as possible to avoid frictions for contributors using the playground and for users trying the README :)


brunoocasali (approved):
LGTM! 🥇

brunoocasali:
bors merge

Issue #642, opened 2022-01-13
Cypress tests consitency


curquiza:
When trying to update this package to be compatible with MeiliSearch v0.25.0, we had some issues with the cypress tests.
We noticed
`Tests / cypress-run` (cf [this file](https://github.com/meilisearch/instant-meilisearch/blob/main/.github/workflows/test.yml)) are not running the same cypress test as `Pre-Release Tests / cypress-run` (cf [this file](https://github.com/meilisearch/instant-meilisearch/blob/main/.github/workflows/pre-release-tests.yml))

`Tests / cypress-run`
- runs using the Vue playground
- runs against a REMOTE MeiliSearch instance (`https://integration-demos.meilisearch.com`)

`Pre-Release Tests / cypress-run`
- runs using the React playground
- runs against a local MeiliSearch instance, downloaded in the CI that will fetch the latest RC

We should
- use the same playground to ensure consistency between the pre-release tests and the others.
- use a local MeiliSearch to run the `Tests / cypress-run` tests. This will avoid update the remote instance to make the tests be green

@bidoubiwa we need your insight before implementing this, you might know why it was done like this. Is there any constraint regarding what I suggest?

bidoubiwa:
I totally agree! See #559 
Tests are only done on the local react server. Tests on other environments where never really done so I removed them. Playgrounds are still running correctly !

PR #659, opened 2022-02-01
Remove tests on remote meilisearch server

bidoubiwa:
fixes: #642 

Since tests were only done on one environment any way during CI but also local testing I removed the testing all together on playground using the remote meilisearch instance


alallema (dismissed):
LGTM! 💪

bidoubiwa:
bors merge

cypress/integration/search-ui.spec.js
```diff
@@ -46,7 +40,7 @@ describe(`${playground} playground test`, () => {
     const select = `.ais-SortBy-select`
     cy.get(select).select('steam-video-games:recommendationCount:asc')
     cy.wait(1000)
-    cy.get(HIT_ITEM_CLASS).eq(0).contains('Rag Doll Kung Fu')
+    cy.get(HIT_ITEM_CLASS).eq(0).contains('Deathmatch Classic')
```
brunoocasali, line 43: Do you understand why this changed between v0.24 and 0.25? I tried to reproduce this behavior to @curquiza but I wasn’t able to :/ 

I saw this happens in the rails integration and here https://github.com/meilisearch/instant-meilisearch/pull/638

brunoocasali (commented):
Just left thought about some questions regarding the change of the results.

cypress/integration/search-ui.spec.js
bidoubiwa: In this specific case it is because I reduced the dataset significantly. Before that, the tests were made against a remote Meilisearch instance containing 10k games. Now there are only 20 games in total. 

Nonetheless, there have been changes between the two versions: https://github.com/meilisearch/instant-meilisearch/pull/615/files

In my case not sure why, there may be changes sometimes between versions so I didn't think to much of it. But it might be important!

cypress/integration/search-ui.spec.js
```diff
@@ -57,28 +51,23 @@ describe(`${playground} playground test`, () => {
...
-
-    cy.contains('Counter-Strike').should('not.exist')
+    cy.wait(1000)
+    cy.get(HIT_ITEM_CLASS).eq(1).contains('Team Fortress Classic')
+    cy.get(HIT_ITEM_CLASS).eq(1).contains('4.99 $')
   })
 
   it('Search', () => {
-    cy.get('.ais-SearchBox-input').type('orwell')
+    cy.get('.ais-SearchBox-input').type('Half-Life')
```
alallema, line 62: I don't think that you were testing search in lower case but the old test was in lower case so just in case.

alallema (approved):
LGTM!
I just add a comment but I don't think it's relevant so I approve it! 🦩

cypress/integration/search-ui.spec.js
brunoocasali: Oh, I see.

Let's keep that in mind, maybe the core team could do something to prevent this in the future :D

.github/workflows/test.yml
```diff
@@ -30,15 +30,23 @@ jobs:
...
+        run: |
+          curl -L https://install.meilisearch.com | sh
+          chmod +x meilisearch
+      - name: Run Meilisearch
+        run: |
+          ./meilisearch --master-key=masterKey --no-analytics true &
       - name: Install dependencies
-        run: yarn --dev && yarn --cwd ./playgrounds/vue
-      - name: Run Browser tests
+        run: yarn --dev && yarn --cwd ./tests/env/react
```
brunoocasali, line 41: Should we create a matrix with react, vue, angular and vanilla javascript? Just in case to prevent possible breaking between them.

bidoubiwa: It would take a very very long time for the tests to run. I thought maybe run them once a week in a scheduled job but not on every PR. What do you think ?

bidoubiwa:
bors merge

PR #670, opened 2022-02-16
Rename test file to be consistent with other repositories

curquiza:
This will make it easy to run scripts in every repository 😇


brunoocasali (approved):
@bidoubiwa I can help you with that! ;)

PR #686, opened 2022-03-14
Bump meilisearch version

brunoocasali:
Update meilisearch.js version to v0.25.0


alallema (approved):
LGTM! But your tests don't seems to be happy

PR #693, opened 2022-03-16
Update Bors badge

curquiza:
Update bors badge link since we use self hosted bors instead of the service


bidoubiwa (approved):
bors merge

PR #710, opened 2022-03-23
Fix placeholdersearch

### LINKED ISSUE #709: placeholderSearch option doesn't work in v0.6.2
**Description**
Setting the configuration property placeholderSearch to false still shows results for me, even though the query is empty. This is contradictory with the documentation.

It used to work correctly in version 0.6.1

**Expected behavior**
Empty query should not trigger a search and not display results.

**Repro**
https://codesandbox.io/s/ms-angularis-forked-cfrjt4?file=/src/app/app.component.ts

bidoubiwa:
# Pull Request

## What does this PR do?
Fixes #709 

`placeholderSearch` is suppose to not showcase any search result when the query is empty. It now correctly does what it is suppose to do. It is also improved as it will now avoid to do a search request to Meilisearch.

### DIFF
--- a/src/adapter/search-request-adapter/__tests__/search-params.tests.ts
+++ b/src/adapter/search-request-adapter/__tests__/search-params.tests.ts
@@ -183,3 +183,29 @@ test('Adapt SearchContext with no finite pagination and pagination total hits lo
 
   expect(searchParams.limit).toBe(4)
 })
+
+test('Adapt SearchContext placeholderSearch set to false', () => {
+  const searchParams = adaptSearchParams({
+    indexUid: 'test',
+    query: '',
+    pagination: { paginationTotalHits: 4, page: 0, hitsPerPage: 6 },
+    defaultFacetDistribution: {},
+    finitePagination: false,
+    placeholderSearch: false,
+  })
+
+  expect(searchParams.limit).toBe(0)
+})
+
+test('Adapt SearchContext placeholderSearch set to false', () => {
+  const searchParams = adaptSearchParams({
+    indexUid: 'test',
+    query: '',
+    pagination: { paginationTotalHits: 200, page: 0, hitsPerPage: 6 },
+    defaultFacetDistribution: {},
+    finitePagination: false,
+    placeholderSearch: true,
+  })
+
+  expect(searchParams.limit).toBe(7)
+})
--- a/src/contexts/search-context.ts
+++ b/src/contexts/search-context.ts
@@ -26,6 +26,7 @@ export function createSearchContext(
     hitsPerPage: instantSearchParams?.hitsPerPage,
     page: instantSearchParams?.page,
   })
+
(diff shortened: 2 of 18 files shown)


alallema (approved):
LGTM! 🌯

bidoubiwa:
bors merge

src/adapter/search-request-adapter/__tests__/search-params.tests.ts
```diff
@@ -183,3 +183,29 @@ test('Adapt SearchContext with no finite pagination and pagination total hits lo
...
+    pagination: { paginationTotalHits: 4, page: 0, hitsPerPage: 6 },
+    defaultFacetDistribution: {},
+    finitePagination: false,
+    placeholderSearch: false,
+  })
+
+  expect(searchParams.limit).toBe(0)
+})
+
+test('Adapt SearchContext placeholderSearch set to false', () => {
```
brunoocasali, line 200: This test has the same description as the first one :)

src/contexts/search-context.ts
```diff
@@ -26,14 +26,15 @@ export function createSearchContext(
...
+
   const searchContext: SearchContext = {
     ...options,
     ...instantSearchParams,
     sort: sortByArray.join(':') || '',
     indexUid,
     pagination,
     defaultFacetDistribution,
-    placeholderSearch: !options.placeholderSearch, // true by default
+    placeholderSearch: options.placeholderSearch !== false, // true by default
```
brunoocasali, line 37: This option is protected by type? Because if the user add `123`, `0` in this option, the option will be `true` :)

tests/placeholder-search.tests.ts
```diff
@@ -0,0 +1,57 @@
...
+    await meilisearchClient
+      .index('movies')
+      .updateFilterableAttributes(['genres'])
+    const documentsTask = await meilisearchClient
+      .index('movies')
+      .addDocuments(dataset)
+    await meilisearchClient.index('movies').waitForTask(documentsTask.uid)
+  })
+
+  test('Test placeholdersearch set to false', async () => {
```
brunoocasali, line 22: What about describing what the test should be expecting?

`Responds with no hits when placeholderSearch is falsy`.

tests/placeholder-search.tests.ts
```diff
@@ -0,0 +1,57 @@
...
+    const response = await customClient.search<Movies>([
+      {
+        indexName: 'movies',
+      },
+    ])
+    const hits = response.results[0].hits
+    expect(hits.length).toBe(5)
+  })
+
+  test('Test placeholdersearch set to true', async () => {
```
brunoocasali, line 40: There is a typo here (the setup use `placeholderSearch: false`)

brunoocasali (commented):
@bidoubiwa i missed the timing here, but I really like to left some notes 🌮

PR #726, opened 2022-04-04
Object constructor for search parameters

bidoubiwa:
In an effort to make a more maintainable code the huge function creating the search parameters is replaced by a more modular object creator.

It is made based on the adapter pattern see [refactoring guru](https://refactoring.guru/design-patterns/adapter/typescript/example). With a somewhat similar example in javascript [here](https://www.dottedsquirrel.com/adapter-pattern-javascript/)

### DIFF
--- a/src/adapter/search-request-adapter/search-params-adapter.ts
+++ b/src/adapter/search-request-adapter/search-params-adapter.ts
@@ -6,6 +6,14 @@ import {
 } from './geo-rules-adapter'
 import { adaptFilters } from './filter-adapter'
 
+/**
+ * Builder to creates an object containing all the search query parameters
+ * provided by the search context of instantsearch.js and instant-meilisearch.
+ *
+ * @param  {SearchContext} searchContext
+ *
+ * @returns {MeiliSearchParams}
+ */
 function ParamsBuilder(searchContext: SearchContext) {
   const meiliSearchParams: Record<string, any> = {}
 
--- a/src/adapter/search-request-adapter/search-params-adapter.ts
+++ b/src/adapter/search-request-adapter/search-params-adapter.ts
@@ -6,6 +6,92 @@ import {
 } from './geo-rules-adapter'
 import { adaptFilters } from './filter-adapter'
 
+function ParamsBuilder(searchContext: SearchContext) {
+  const meiliSearchParams: Record<string, any> = {}
+
+  return {
+    getParams() {
+      return meiliSearchParams
+    },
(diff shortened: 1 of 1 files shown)


src/adapter/search-request-adapter/search-params-adapter.ts
```diff
@@ -6,6 +6,100 @@ import {
...
+ * to meilisearch search query parameters.
+ *
+ * @param  {SearchContext} searchContext
+ *
+ * @returns {MeiliSearchParams}
+ */
+function ParamsAdapter(searchContext: SearchContext) {
+  const meiliSearchParams: Record<string, any> = {}
+
+  return {
```
brunoocasali, line 20: Entering the real question now, I think we should check the "creational patterns", more specifically the builder https://refactoring.guru/design-patterns/builder.

But in a way, we can fit the functional style we are looking to develop our JS apps.
To express this idea, we could rename the functions to `buildStuff` or even better `addStuff`, since we are actually adding properties to the object, what do you think about it? 🧥 

I tried some different versions of your implementation, trying to combine `builder`+`imutability` (basically I'm trying to implement what this guy [says here](https://youtu.be/0if71HOyVjY?t=1088)).

But neither of my iterations was OK. 

I came up with a solution which looks like this (but I'm still not satisfied with it):

```
  const builder = ParamsBuilder(searchContext)
    .addFacets()
    .addFilters()
    .addLimit()
    .addSort()
    .addGeoSearchRules()
    .addAttributesToCrop()
    .addAttributesToHighlight()
    .addAttributesToRetrieve()

  return builder.build()
```

And my initial idea of `meiliParams = addFilters(context, meiliParams)` does not work very well because we still have to mutate the object inside of it.

But the implementation relies on the `this` and is still mutating data (I had some trouble with the typescript `type` thing). 

I found some resources to help implement this:

- https://javascript.plainenglish.io/implementing-the-builder-pattern-in-javascript-without-classes-eaf41f93b9c0
- https://medium.com/front-end-weekly/how-to-use-powerful-function-composition-in-javascript-3c0447642044

brunoocasali (commented):
@bidoubiwa, I need to understand how typescript works and how we can deal with a _legacy_ OOP/Class oriented style to functional programming.

In general just the naming I think we can iterate over 😄 

I left some comments about the code, but in the end, I want to say that is precisely the kind of discussion I would like to have here :)

Congrats on your work!

brunoocasali (approved):
❤️

PR #738, opened 2022-04-21
Add crop marker support

bidoubiwa:
# Pull Request

## What does this PR do?
Previously instant-meilisearch added crop markers in the specific fields depending on the `snippetEllipsisText` value provided by the `configuration` widget of instantsearch. The injection of the markers was done by instant-meilisearch. Now it is added as the query parameter `cropMarker` to the search request where the markers are added by meilisearch the engine.

Some tests have been removed as they where not relevant.


alallema (approved):
LGTM! 🌯

PR #739, opened 2022-04-21
Add highlight prefix suffix

bidoubiwa:
# Pull Request

## What does this PR do?
Fixes #<issue_number>

## PR checklist
- [ ] Does this PR fix an existing issue?
- [ ] Have you read the contributing guidelines?
- [ ] Have you made sure that the title is accurate and descriptive of the changes?


tests/highlight.tests.ts
```diff
@@ -231,13 +231,18 @@ describe('Highlight Browser test', () => {
...
         },
       },
     ])
+
     const hit = response.results[0].hits[0]._highlightResult
 
+    if (hit?.title) {
+      expect(hit?.title?.value).toEqual('Ariel')
+    }
+
```
alallema, line 245: I don't understand how you get this result with your query?

src/adapter/search-request-adapter/search-params-adapter.ts
```diff
@@ -63,6 +63,22 @@ export function adaptSearchParams(
     '*',
   ]
 
+  // Highlight pre tag
+  const highlightPreTag = searchContext?.highlightPreTag
+  if (highlightPreTag) {
+    meiliSearchParams.highlightPreTag = highlightPreTag
+  } else {
+    meiliSearchParams.highlightPreTag = '__ais-highlight__'
```
alallema, line 71: `__ais-highlight__` is the default highlight of instant-search?

alallema (commented):
Well done! Just ask few questions 🌮

src/adapter/search-request-adapter/search-params-adapter.ts
bidoubiwa: Absolutely! It is actually a bad idea to change this in instant-meilisearch as the tags are not rendered as HTML. 

Screenshot 2022-04-25 at 12 23 23

So the user should stick with the default value. 
I added it because the user might want to add other things than tags and also to be consistent with our search parameters.

alallema (approved):
LGTM! 🍘

PR #780, opened 2022-06-29
Pass client agent to meilisearch-js

bidoubiwa:
See [related issue](https://github.com/meilisearch/integration-guides/issues/150)

This PR passes down the information of the instant-meilisearch package to the meilisearch-js package which in turn provided the information to meilisearch.

### DIFF
--- a/.github/scripts/check-release.sh
+++ b/.github/scripts/check-release.sh
@@ -2,17 +2,16 @@
 
 # Checking if current tag matches the package version
 current_tag=$(echo $GITHUB_REF | cut -d '/' -f 3 | tr -d ' ',v)
-file_tag=$(grep '"version":' package.json | cut -d ':' -f 2- | tr -d ' ' | tr -d '"' | tr -d ',')
-if [ "$current_tag" != "$file_tag" ]; then
-  echo "Error: the current tag does not match the version in package file(s)."
-  echo "$current_tag vs $file_tag"
-  exit 1
-fi
 
-package_version_ts=$(grep "PACKAGE_VERSION =" src/package-version.ts | cut -d "=" -f 2- | tr -d " " | tr -d "'")
-if [ "$current_tag" != "$package_version_ts" ]; then
-  echo "Error: the current tag does not match the version in src/package-version.ts."
-  echo "$current_tag vs $package_version_ts"
+package_file_tag=$(grep '"version":' package.json | cut -d ':' -f 2- | tr -d ' ' | tr -d '"' | tr -d ',')
+package_file_name="package.json"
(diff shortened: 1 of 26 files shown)


.github/scripts/check-release.sh
```diff
@@ -9,5 +9,12 @@ if [ "$current_tag" != "$file_tag" ]; then
   exit 1
 fi
 
+package_version_ts=$(grep "PACKAGE_VERSION =" src/package-version.ts | cut -d "=" -f 2- | tr -d " " | tr -d "'")
+if [ "$current_tag" != "$package_version_ts" ]; then
+  echo "Error: the current tag does not match the version in src/package-version.ts."
+  echo "$current_tag vs $package_version_ts"
+  exit 1
+fi
```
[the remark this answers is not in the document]
curquiza: What we do in the Meilisearch repo. Don't know if it's exactly the same usecase, but it can maybe help: https://github.com/meilisearch/meilisearch/blob/release-v0.28.0/.github/scripts/check-release.sh

.github/scripts/check-release.sh
```diff
@@ -2,10 +2,16 @@
...
-if [ "$current_tag" != "$file_tag" ]; then
-  echo "Error: the current tag does not match the version in package file(s)."
-  echo "$current_tag vs $file_tag"
+
+package_file_tag=$(grep '"version":' package.json | cut -d ':' -f 2- | tr -d ' ' | tr -d '"' | tr -d ',')
+package_file_name='package.json'
+version_file_tag=$(grep "PACKAGE_VERSION =" src/package-version.ts | cut -d "=" -f 2- | tr -d " " | tr -d "'")
+version_file_name='src/package-version.ts'
+
+if [ "$current_tag" != "$package_file_tag" ] || [ "$current_tag" != "$version_file_tag"]; then
```
curquiza, line 11:
```suggestion
if [ "$current_tag" != "$package_file_tag" ] || [ "$current_tag" != "$version_file_tag" ]; then
```

curquiza: Capture d’écran 2022-06-30 à 13 36 24

curquiza (changes requested):

.github/scripts/check-release.sh
bidoubiwa: Weird, it works for me
```
Error: the current tag does not match the version in package file(s).
package.json: 0.28.0 vs 0.7.1
src/package-version.ts: 0.28.0 vs 0.7.1
```

curquiza: My example was checking in the case everything is working (with `GITHUB_REF=refs/tags/v0.7.1`)

curquiza (approved):

PR #794, opened 2022-07-16
Update search key in playgrounds

bidoubiwa:
With the upgrade of versions in the [integration Meilisearch instance](https://integration-demos.meilisearch.com) the keys have been regenerated and thus are not working anymore.

This time the key has been created with a specific uid. On next updates of the Meilisearch server instead of changing the API key we just need to recreate the key in the new instance and the key will still be working.


brunoocasali (approved):
🎉

PR #798, opened 2022-07-20
Add beta pagination feature

bidoubiwa:
We would like users to be able to test out our new way to handle pagination before adding it to Meilisearch.

While previously if the user wanted to have a pagination with page number (Ex: `<< < 1, 2, 3, ... > >>`) they had to use `finitePagination:true` and `paginationTotalHits`. 

These have been removed as Meilisearch now handles this type of pagination.

The following parameters in the `instantMeiliSearch` client are removed: 
- `finitePagination`
- `paginationTotalHits`

See [related issue](https://github.com/meilisearch/meilisearch/issues/2577)
See [related discussion](https://github.com/meilisearch/product/discussions/483)


tests/env/react/src/App.js
```diff
@@ -55,13 +54,13 @@ const App = () => (
...
         <h2>Players</h2>
-        <RefinementList attribute="players" />
+        <RefinementList attribute="players" operator="or" />
         <h2>Platforms</h2>
         <RefinementList attribute="platforms" />
         <h2>Misc</h2>
         <RefinementList attribute="misc" />
         <Configure
-          hitsPerPage={6}
+          hitsPerPage={1}
```
bidoubiwa, line 63:
```suggestion
          hitsPerPage={6}
```

src/adapter/search-response-adapter/__tests__/pagination-adapter.tests.ts
```diff
@@ -42,85 +47,133 @@ const paginateHitsTestsParameters = [
...
+    },
   },
 
   // Page 0
   {
     hits: [{ id: 1 }, { id: 2 }, { id: 3 }],
     page: 0,
     hitsPerPage: 20,
-    returnedHits: [{ id: 1 }, { id: 2 }, { id: 3 }],
+    returnedPagination: { page: 0, hitsPerPage: 20, nbPages: 1 },
```
brunoocasali, line 82: I think you can opt to use multi-line or one-line but as soon you opt by one version you must keep the consistency :)

src/adapter/search-response-adapter/total-hits-adapter.ts
```diff
@@ -0,0 +1,18 @@
...
+    totalPages = 0,
+    estimatedTotalHits = undefined,
+    totalHits = undefined,
+  } = searchResponse
+  if (estimatedTotalHits !== undefined) {
+    return estimatedTotalHits
+  } else if (totalHits !== undefined) {
+    return totalHits
+  }
+  return hitsPerPage * totalPages
```
brunoocasali, line 17: According to our internal conversation, this will change in the future, right @bidoubiwa ?

bidoubiwa: > According to our internal conversation, this will change in the future,

What do you mean?

brunoocasali: The idea here is that the core team provides a real number of hits we have. Because even we doing a math here we cannot be 100% sure since we may have pages without the whole limit.

bidoubiwa: Its the if just above 

```
else if (totalHits !== undefined) {
    return totalHits
  }
```

Thats the hit Meilisearch returns. The return is a safe guard in case total hits is missing but it should indeed not happen so I could as well remove it

bidoubiwa:
It has been implemented in #878 and released :)

PR #801, opened 2022-07-21
Create new publish CI for beta versions

bidoubiwa:
To be able to create beta releases on different subjects: `bumps`, `features`, etc.. we needed to provide a better suited CI.

This CI also checks if the tag well formated to ensure we do not publish a version that does not follow the naming convention.

The workflow has been tested on a private directory and seems to work correctly.


brunoocasali (approved):
🙈 🎉

bidoubiwa:
bors merge

PR #808, opened 2022-07-25
Add ability to clear cache to allow real-time refresh of search results.

### LINKED ISSUE #790: Invalidate client cache
**Description**

Currently meilisearch's client allows preserving the facets with count of 0 thanks to the `keepZeroFacets` option which can be provided to the constructor of the client. Looks like the client is using a caching system in order to achieve that. 

Sometimes this results in unexpected behavior. For example, if I have page A, which displays a list of categories, and when a user clicks on that category the user is redirected to page B where more refinements could be provided based on it.
For this example, let's say a user clicks on the "Shoes" category, and after he is redirected to page B he would also like to refine the results by size. Normally you would expect to see sizes related to the "Shoes" category, such as "38", "39", "40", and if you are searching for "T-Shirts" you expect sizes such as "S", "M", "L".
The way I achieve this is using the 
```jsx 
<VirtualMenu attribute="category" defaultRefinement={parameterFromPageA} />
```
 component on page B which on mount refines the search with the parameter that you provided from page A (in this case "Shoes"). 

The problem is that the facets that you get for sizes are cached, and once they are cached if you change your category to T-Shirts you would also see the sizes for Shoes, which is counterintuitive for a user. 
For now, I solved this problem by creating a new client initially when page B loads, but I am not happy with this. a better solution would be if you could somehow invalidate the current cache.

This is what my current solution looks like:

```ts
const searchClient = useMemo(createMeiliSearchClient, [])
```
**Basic example**

The way I would use this new feature would be as following:

```ts
useEffect(() => {
  searchClient.invalidateCache()
}, [])
```

**Other**
I would also be happy if you share your opinion on the way I achieve this result and if there is an intended way of doing such stuff by the API which I am missing.

fadeaway:
# Pull Request

## What does this PR do?
Fixes [#790](https://github.com/meilisearch/instant-meilisearch/issues/790)

Implements clearCache method on InstantMeiliSearchInstance to allow realtime updates to search results. This allows algolias refresh example to work out of the box. ([https://www.algolia.com/doc/guides/building-search-ui/going-further/improve-performance/vue/#caching](https://www.algolia.com/doc/guides/building-search-ui/going-further/improve-performance/vue/#caching))

## PR checklist
- [X ] Does this PR fix an existing issue?
- [X ] Have you read the contributing guidelines?
- [X ] Have you made sure that the title is accurate and descriptive of the changes?

### DIFF
--- a/src/client/instant-meilisearch-client.ts
+++ b/src/client/instant-meilisearch-client.ts
@@ -1,20 +1,19 @@
+import { MeiliSearch } from 'meilisearch'
 import {
-  AlgoliaMultipleQueriesQuery,
-  AlgoliaSearchResponse,
-  InstantMeiliSearchInstance,
   InstantMeiliSearchOptions,
+  InstantMeiliSearchInstance,
+  AlgoliaSearchResponse,
+  AlgoliaMultipleQueriesQuery,
   SearchContext,
 } from '../types'
-import { SearchCache, cacheFirstFacetDistribution } from '../cache/'
 import {
-  SearchResolver,
-  adaptSearchParams,
   adaptSearchResponse,
+  adaptSearchParams,
+  SearchResolver,
 } from '../adapter'
-
-import { MeiliSearch } from 'meilisearch'
-import { constructClientAgents } from './agents'
 import { createSearchContext } from '../contexts'
+import { SearchCache, cacheFirstFacetDistribution } from '../cache/'
+import { constructClientAgents } from './agents'
 
 /**
  * Instanciate SearchClient required by instantsearch.js.
--- a/src/client/instant-meilisearch-client.ts
+++ b/src/client/instant-meilisearch-client.ts
@@ -1,19 +1,20 @@
-import { MeiliSearch } from 'meilisearch'
 import {
-  InstantMeiliSearchOptions,
-  InstantMeiliSearchInstance,
-  AlgoliaSearchResponse,
   AlgoliaMultipleQueriesQuery,
+  AlgoliaSearchResponse,
+  InstantMeiliSearchInstance,
+  InstantMeiliSearchOptions,
   SearchContext,
 } from '../types'
-import {
-  adaptSearchResponse,
-  adaptSearchParams,
-  SearchResolver,
-} from '../adapter'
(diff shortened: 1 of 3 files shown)


src/client/instant-meilisearch-client.ts
```diff
@@ -1,19 +1,20 @@
...
   SearchResolver,
+  adaptSearchParams,
+  adaptSearchResponse,
 } from '../adapter'
-import { createSearchContext } from '../contexts'
-import { SearchCache, cacheFirstFacetDistribution } from '../cache/'
+
+import { MeiliSearch } from 'meilisearch'
 import { constructClientAgents } from './agents'
+import { createSearchContext } from '../contexts'
```
bidoubiwa, line 17: I don't the order of import needs to be changed, could you rollback?

src/types/types.ts
```diff
@@ -1,9 +1,10 @@
 import type {
-  SearchResponse as MeiliSearchResponse,
   FacetDistribution,
+  SearchResponse as MeiliSearchResponse,
 } from 'meilisearch'
-import type { SearchClient } from 'instantsearch.js'
+
 import type { MultipleQueriesQuery as AlgoliaMultipleQueriesQuery } from '@algolia/client-search'
+import type { SearchClient } from 'instantsearch.js'
```
bidoubiwa, line 7: I don't the order of import needs to be changed, could you rollback?

bidoubiwa (changes requested):
A part from the order of the imports it looks very good to me :)

src/client/instant-meilisearch-client.ts
fadeaway: The import order has been reverted.

src/types/types.ts
fadeaway: The import order has been reverted.

bidoubiwa (approved):
LGTM 🔥 thanks a lot :)

bidoubiwa:
bors merge

dasraab:
i think this feature is missing in the official documentation. would be great to add that.

bidoubiwa:
#893 :) thanks!

PR #811, opened 2022-07-26
Use needs context in beta CI to communicate meilisearch image version

bidoubiwa:
To communicate custom information to different jobs github recommends using the `needs` context. 
[See documentation](https://docs.github.com/en/actions/learn-github-actions/contexts#example-usage-of-the-needs-context)

In this CI, we are using cypress which does not have access to the repository files. Since we need access to it to be able to fetch the version of the docker image, we use this method as suggested by them [See doc](https://github.com/cypress-io/github-action#robust-custom-build-id).


brunoocasali (approved):

bidoubiwa:
bors merge

PR #815, opened 2022-08-01
Update nodeJs requirements in readme and CI

bidoubiwa:
Fixes the node js compatibility in readme
and make publishing use node 16


alallema (approved):
LGTM! 🔥

PR #816, opened 2022-08-01
Update beta guide in contributing

bidoubiwa:
The beta guide was wrongly telling the user to create a specific formated branch which is not the one required.


alallema (approved):
Great! LGTM! ✨

PR #817, opened 2022-08-02
Feature prototype: define strategy on optional words

bidoubiwa:
Introduces `optionalWords` search parameter.
It lets you decide the search strategy for words matching during a search.

See https://github.com/meilisearch/meilisearch/pull/2636

`optionalWords` gives you the possibility to chose how Meilisearch should handle the presence of multiple query words.

For example, if your query is `Hello world` by default Meilisearch returns documents containing either both `Hello` and `world` or documents that only contain `hello`. This is the `last` strategy, where words are stripped from the right.
The other strategy is `none`, where both `hello` and `worlds` **must** be present in a document for it to be returned.

```js
import { instantMeiliSearch } from '@meilisearch/instant-meilisearch'

  host,
  apiKey,
  {
      optionalWords: 'none' // default last
  }
)
```


gmourier (approved):

brunoocasali (approved):

bidoubiwa:
included in #831

PR #818, opened 2022-08-02
Remove unecessary clientParams type

bidoubiwa:
the type `ClientParams` was used nowhere and added a useless complexity to the types.

`paginationTotalHits` is not added in the `SearchContext` as it is added in, and used from the `PaginationContext` 

Types are also re-ordered in order to get the mandatory fields at the top and the optional after them


brunoocasali (approved):

bidoubiwa:
bors merge

Issue #797, opened 2022-07-20
keepZeroFacets interaction with cache when pre-populating results does not work


andykirk12:
**Description**
When keepZeroFacets: true, and we  pre-select facets in the refinementList widget - either through the routing method, or by manually setting the value with refinementlist.refine() - we are seeing that the zero-count facets are not being made available. This is not true when we perform the same search manually by clicking the facets from an empty search state.

_Manual Search:_
![image](https://user-images.githubusercontent.com/109536123/179984978-9abb19a2-e9f3-42b1-9fa5-c78633651455.png)

_Pre-population from URL or refinementList.refine():_
![image](https://user-images.githubusercontent.com/109536123/179985097-4e88310a-b7de-42bb-bb02-b4a66f658a83.png)

_And then unselect the pre-populated facet, then select a different facet in the same refinementList:_
![image](https://user-images.githubusercontent.com/109536123/179985876-d54b4c0a-3900-4a57-8c4d-34fc2a5c3311.png)

**What I believe is happening**
I believe the issue is based in the way the facets cache is implemented. As far as I can tell, the cache is initially loaded when the search begins, and this is what gives us the "default" list of facets in the cache. The library then adds/removes from that to generate the list. The issue comes when we move into a more "unrefined" search state, the cache isn't updated to reflect that.

 **Suggestion**
I think that a good way to resolve this is to give a method to clear the cache state, or refresh it to the current search criteria. It appears that the InstantSearch.js method of search.refresh() does not clear out the instant-meilisearch cache, as described in their docs.

Issue #825, opened 2022-08-22
keepZeroFacets does not work with routing


ar-siddiqui:
**Description**
keepZeroFacets does not work in instant-meilisearch react correctly 

**Expected behavior**
return 0 facets as well

**Current behavior**
does not return 0 facets

Tried both with and without `placeHoloderSearch`

bidoubiwa:
For history 

> In our use case, we do not like the meilisearch API not returning 0 facets. We want to show users to see 0 value facets as well so that they can select them. The hack provided by this [PR](https://github.com/meilisearch/instant-meilisearch/pull/695) only works when you first do a no filter search, if you directly use a URL with specific rules it would not showcase them as the client does not have anything to remember from.

> IMO the meilisearch API should always return 0 facets (or be configurable to return 0 facets) and the client could be configured to either display them or not.

> For example, displaying 0 facets allow a user to add `spreadsheet` as well to the search
![image](https://user-images.githubusercontent.com/53625184/159548177-fb5892da-2730-4548-8c61-55394af5d28c.png)

> But without 0 facets, the user is not even aware that a type `spreadsheet` exists.
> ![image](https://user-images.githubusercontent.com/53625184/159548495-135711aa-ddd7-43d3-b9f1-9fc7f7c1f358.png)

bidoubiwa:
The issue appears in two circumstances:
- When `keepzerofacets` is used with `placeholdersearch: false` (which is not the case here)
- When `keepzerofacets` is used with rooting (which is the case here)

ar-siddiqui:
The issue is there even without routing

bidoubiwa:
In your quoted answer in my comment you are stating 

> directly use a URL with specific rules it would not showcase them as the client does not have anything to remember from.

Could you update your issue with more context to be able to understand the issue?

bidoubiwa:
Hey @ar-siddiqui could you try your bug again with the latest release ? (v0.8.2)

ar-siddiqui:
> Hey @ar-siddiqui could you try your bug again with the latest release ? (v0.8.2)

Ok, I can confirm that this is working and `zerofacets` are displayed, but there is another issue. I will open a new issue to track that.

ar-siddiqui:
> > Hey @ar-siddiqui could you try your bug again with the latest release ? (v0.8.2)
> 
> Ok, I can confirm that this is working and `zerofacets` are displayed, but there is another issue. I will open a new issue to track that.

[...]

Issue #776, opened 2022-06-23
keepZeroFacets not keeping facets with 0 matches when using custom search method with search client


impzero:
**Description**

Hello, I am trying to preserve the facets from my RefinementList when I select one, but the ones which are not selected disappear.
Let's say we have 3 categories - **clothes**, **shoes**, and **accessories**, clicking on "**clothes**" would hide **shoes** and **accessories** until I unselect **clothes**
I saw in the README that you could provide a `keepZeroFacets` property to the client since version `0.6.2`
I updated my client, set the property to true, and also made sure that my attributes are filterable in the backend. Still, I can't retain the non-chosen categories in the RefinementList

**Expected behavior**

As the documentation from [README ](https://github.com/meilisearch/instant-meilisearch#-customization)states, I expected that after a refinement, facets/filters with 0 matches would still appear in the list

**Current behavior**

As described, currently when refinement finishes, my RefinementList the `items` prop contains only the currentRefinement (for example the selected category - **clothes)**

**Steps to reproduce**

Thankfully I managed to reproduce this in a code sandbox [here](https://codesandbox.io/s/vigilant-brook-sszorg)

**Environment**
- OS: EndeavourOS
- Browser: Chrome Version 102.0.5005.115 (Official Build) (64-bit)
- Meilisearch version: 0.26.1
- instant-meilisearch version: 0.6.2

Sidran83:
We have the same problem, the ulselected facets disappear when one is selected in our efinementList. Is there any fix yet?

bidoubiwa:
Hey! I think this might be linked to #775 but I'm not sure! I will investigate this issue after the pre-release of Meilisearch v0.28.0 ( after 11 july)

bidoubiwa:
Are you using react by any chance?

ar-siddiqui:
I am using react and no matter what I try I can't get 0 facets.

> Are you using react by any chance?

bidoubiwa:
Thanks @ar-siddiqui there seems to be an issue with how React understands the conditionals of facets.

bidoubiwa:
Should be fixed! If it is not, please tell me and re open :)

PR #830, opened 2022-08-29
Fix default zero facets on placeholdersearch and routing

bidoubiwa:
fixes: #797 

Maybe related: #776
Maybe related: #825 

`keepZeroFacets` was not working when `placeholderSearch` was set to false and when using the routing system. 

This PR fixes the issue.

### DIFF
--- a/tests/first-facets-distribution.tests.ts
+++ b/tests/first-facets-distribution.tests.ts
@@ -148,7 +148,7 @@ describe('Default facet distribution', () => {
   // with keepZeroFacets
   // with placeholderSearch
   // With multiple search (empty and no results expected)
-  test('creation of facet distribution with facets and keepZeroFacets to true', async () => {
+  test('Ensure cached facetDistribution works between two searches when keepZeroFacets is true', async () => {
     const searchClient = instantMeiliSearch(HOST, API_KEY, {
       keepZeroFacets: true,
     })
@@ -202,7 +202,7 @@ describe('Default facet distribution', () => {
   // with keepZeroFacets
   // without placeholderSearch
   // With multiple search (empty and no results expected)
-  test('creation of facet distribution with facets, keepZeroFacets to true and placeholderSearch to false', async () => {
+  test('ensure cached facetDistribution works between two calls when placeholderSearch is false', async () => {
     const searchClient = instantMeiliSearch(HOST, API_KEY, {
       keepZeroFacets: true,
       placeholderSearch: false,


tests/first-facets-distribution.tests.ts
```diff
@@ -0,0 +1,148 @@
+import { dataset, meilisearchClient, HOST, API_KEY } from './assets/utils'
+import { instantMeiliSearch } from '../src'
+
+describe('Firt facet distribution', () => {
```
brunoocasali, line 4:
```suggestion
describe('First facet distribution', () => {
```

tests/first-facets-distribution.tests.ts
```diff
@@ -0,0 +1,148 @@
...
+        Animation: 0,
+        Comedy: 0,
+        Crime: 0,
+        Drama: 0,
+        'Science Fiction': 0,
+        Thriller: 0,
+      },
+    })
+  })
+
```
brunoocasali, line 121: I think you can remove this, because you have the same setup/exercise in the test below.

tests/search-resolver.tests.ts
```diff
@@ -104,7 +104,7 @@ describe('Pagination browser test', () => {
       apiKey: '',
       clientAgents: [`Meilisearch instant-meilisearch (v${PACKAGE_VERSION})`],
     })
-    expect(mockedSearch).toHaveBeenCalledTimes(2)
+    expect(mockedSearch).toHaveBeenCalledTimes(3)
```
brunoocasali, line 107: Why do we have +1 calls now?

tests/assets/utils.ts
```diff
@@ -1,6 +1,9 @@
 import { instantMeiliSearch } from '../../src'
 import { MeiliSearch } from 'meilisearch'
 
+const HOST = 'http://localhost:7700'
```
brunoocasali, line 4: Cosmetic change? 💅

src/client/instant-meilisearch-client.ts
```diff
@@ -43,6 +38,12 @@ export function instantMeiliSearch(
     clientAgents,
   })
 
+  const searchCache = SearchCache()
+  // create search resolver with included cache
+  const searchResolver = SearchResolver(meilisearchClient, searchCache)
+
+  let defaultFacetDistribution: any
```
brunoocasali, line 45: Can `any` in typescript be considered bad practice?

src/cache/first-facets-distribution.ts
```diff
@@ -1,14 +1,20 @@
...
+  searchContext: SearchContext
+): Promise<FacetDistribution> {
+  const meilisearchParams = MeiliParamsCreator(searchContext)
+  meilisearchParams.addFacets()
+  meilisearchParams.addAttributesToRetrieve()
+
+  // Search response from Meilisearch
+  const searchResponse = await searchResolver.searchResponse(
+    // placeholdersearch true to ensure a request is made
+    // query set to empty to ensure default facetdistributionap
```
brunoocasali, line 15: facetdistributionap?

src/cache/first-facets-distribution.ts
```diff
@@ -1,14 +1,20 @@
...
-    return searchResponse.facetDistribution
-  }
-  return defaultFacetDistribution
+export async function cacheFirstFacetDistribution(
+  searchResolver: any,
+  searchContext: SearchContext
+): Promise<FacetDistribution> {
+  const meilisearchParams = MeiliParamsCreator(searchContext)
+  meilisearchParams.addFacets()
+  meilisearchParams.addAttributesToRetrieve()
```
brunoocasali, line 10: Which attributesToRetrieve this query will receive? All of it? If yes, we could try to receive no attributes at all, since we just want the facets... This will reduce the amount of data handled by the client side.

tests/first-facets-distribution.tests.ts
```diff
@@ -0,0 +1,148 @@
...
+        },
+      },
+    ])
+    expect(response.results[0].facets).toEqual({})
+  })
+
+  test('creation of facet distribution without facets and with keepZeroFacets to true', async () => {
+    const searchClient = instantMeiliSearch(HOST, API_KEY, {
+      keepZeroFacets: false,
+    })
```
mdubus, line 34: `with keepZeroFacets to true` in the title of the test, but `keepZeroFacets: false` in the code 👀

tests/first-facets-distribution.tests.ts
```diff
@@ -0,0 +1,148 @@
...
+      {
+        indexName: 'movies',
+        params: {
+          facets: [],
+          query: '',
+        },
+      },
+    ])
+    expect(response.results[0].facets).toEqual({})
+  })
```
mdubus, line 45: Looks like this test is exactly the same as above. It may be related to my previous comment 🙈

mdubus (changes requested):

tests/search-resolver.tests.ts
bidoubiwa: Because the first call always sets up the default facet distribution by making a call with zero filters (except for the facet field) to ensure we have the facet distribution of an empty search not impacted by any filter.

tests/assets/utils.ts
bidoubiwa: Haha, yes I was annoyed that it was hardcoded bellow. My bad if this is not the right PR for this change

src/client/instant-meilisearch-client.ts
bidoubiwa: It can but it's not rare to see it in some places where it is hard to let typescript infer the type. But I actually tried it again and compared to before it doesn't seem to scream anymore so I'll update it!

tests/first-facets-distribution.tests.ts
bidoubiwa: There is actually a difference for me! 
The first tests: 
- Makes a search with no query
- Should cache the default distribution correctly
- Makes a search with the same `searchClient` as before but with a query that would return 0 documents
- Checks if every facet is correctly set to 0

Second test:
- Does a search with a query that would return 0 documents
- Should cache the default distribution when none has been set before
- Checks if every facet is correctly set to 0

There is a cache system between the multiple calls on the same search client. These two tests ensure they are helping each other. 
The tests that ensure that when doing two call on my search client only triggers the default facetdistribution call only once is tested here: https://github.com/meilisearch/instant-meilisearch/pull/830#discussion_r958622286. In this test we are calling two time `searchClient.search`. During the first `search` we make two requests to meilisearch, one for the default facet distribution, one for the actual search. During the second `search` on the same `searchClient` the default facet distribution should already be cached and only the actual search request should be made, thus 3 calls.

tests/search-resolver.tests.ts
bidoubiwa: See https://github.com/meilisearch/instant-meilisearch/pull/830#discussion_r959370174

tests/first-facets-distribution.tests.ts
bidoubiwa: The search client is different

First test
```
    const searchClient = instantMeiliSearch(HOST, API_KEY)
```

Second test:
```
const searchClient = instantMeiliSearch(HOST, API_KEY, {
      keepZeroFacets: false,
    })
```

It checks wether the facet distribution is correctly fetched and returned in both cases

bidoubiwa (commented):
These are the scenario tested in the tests: 

```
  // Without facets
  // Without keepZeroFacets
  // With placeholderSearch
  // With empty query
```

```
// Without facets
  // With keepZeroFacets
  // With placeholderSearch
  // With empty query
```

```
  // With facets
  // without keepZeroFacets
  // without placeholderSearch
  // With multiple search (empty and no results expected)
```

```
  // With facets
  // without keepZeroFacets
  // with placeholderSearch
  // With empty query
```

```
  // With facets
  // with keepZeroFacets
  // with placeholderSearch
  // With multiple search (empty and no results expected)
```

```
  // With facets
  // with keepZeroFacets
  // without placeholderSearch
  // With multiple search (empty and no results expected)
```


```
  // With facets
  // with keepZeroFacets
  // with placeholderSearch
  // Without multiple search
  // Without query expecting no results
```

mdubus (approved):
LGTM 🔥

brunoocasali (approved):
💪

brunoocasali (approved):

bidoubiwa:
bors merge

PR #832, opened 2022-09-01
Update search parameters for v0.29.0

bidoubiwa:
Introduces `optionalWords` search parameter.
It lets you decide the search strategy for words matching during a search.

`matchingStrategy` gives you the possibility to chose how Meilisearch should handle the presence of multiple query words.

For example, if your query is `Hello world` by default Meilisearch returns documents containing either both `Hello` and `world` or documents that only contain `hello`. This is the `last` strategy, where words are stripped from the right.
The other strategy is `all`, where both `hello` and `worlds` **must** be present in a document for it to be returned.

```js
import { instantMeiliSearch } from '@meilisearch/instant-meilisearch'

  host,
  apiKey,
  {
      matchingStrategy: 'all' // default last
  }
)
```

### DIFF
--- a/src/adapter/search-request-adapter/__tests__/search-params.tests.ts
+++ b/src/adapter/search-request-adapter/__tests__/search-params.tests.ts
@@ -1,16 +1,17 @@
 import { adaptSearchParams } from '../search-params-adapter'
+import { MatchingStrategies } from '../../../types'
 
 const DEFAULT_CONTEXT = {
   indexUid: 'test',
   pagination: { paginationTotalHits: 20, page: 0, hitsPerPage: 6 },
   defaultFacetDistribution: {},
+  finitePagination: false,
 }
 
 describe('Parameters adapter', () => {
   test('adapting a basic searchContext ', () => {
     const searchParams = adaptSearchParams({
       ...DEFAULT_CONTEXT,
-      finitePagination: false,
     })
 
     expect(searchParams.attributesToHighlight).toContain('*')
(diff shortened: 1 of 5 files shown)


src/adapter/search-request-adapter/__tests__/search-params.tests.ts
```diff
@@ -162,7 +164,6 @@ describe('Pagination adapter', () => {
     const searchParams = adaptSearchParams({
       ...DEFAULT_CONTEXT,
       pagination: { paginationTotalHits: 4, page: 0, hitsPerPage: 6 },
-      finitePagination: false,
```
brunoocasali, line 165: Can you explain that?

src/contexts/search-context.ts
```diff
@@ -37,6 +38,7 @@ export function createSearchContext(
     placeholderSearch: options.placeholderSearch !== false, // true by default
     keepZeroFacets: !!options.keepZeroFacets, // false by default
     finitePagination: !!options.finitePagination, // false by default
+    matchingStrategy: options.matchingStrategy || MatchingStrategies.LAST,
```
brunoocasali, line 41: The default behavior can be handled by Meilisearch, why do we need to set it here?

src/adapter/search-request-adapter/__tests__/search-params.tests.ts
bidoubiwa: Not very much related to the PR, I wanted to remove these unnecessary lines.

src/contexts/search-context.ts
bidoubiwa: Mostly safeguarding, but indeed not necessary. Would you like me to remove it?

brunoocasali: Just thinking, if this is good or not, if in the future meilisearch decides to change the default value, we will have differences between SDKs...

mdubus (approved):
LGTM 🥳

PR #833, opened 2022-09-01
Change order of branches conditions in beta CI

bidoubiwa:
The order made that the second condition was overriding the first one. Making the tests run anyway on bump beta;s


brunoocasali (approved):

bidoubiwa:
bors merge

PR #878, opened 2022-11-14
Add pagination feature for v0.30.0

bidoubiwa:
Add a new pagination strategy when using the numbered (finite) pagination. 

In v0.30.0, the `pagination` using `page` and `hitsPerPage` is introduced. Using this system, we get rid of `paginationTotalHits` and we now ensure an exhaustive total number of hits that is not capped at `200` or at the value specified in `paginationTotalHits`

To enable it, you still need to use `finitePagination` in your instantMeilisearch settings.


README.md
```diff
@@ -927,7 +909,7 @@ instantsearch.widgets.clearRefinements({
 
 The `pagination` widget displays a pagination system allowing the user to change the current page.
 
-We do not recommend using this widget as pagination slows the search responses. Instead, the [InfiniteHits](#-infinitehits) component is recommended.
+// TODO: explain what is needed for it to work
```
mdubus, line 912: 👀

src/contexts/pagination-context.ts
```diff
@@ -1,20 +1,21 @@
...
+  finite?: boolean,
+  hitsPerPage?: number,
+  page?: number
+): PaginationState {
   return {
-    paginationTotalHits:
-      paginationTotalHits != null ? paginationTotalHits : 200,
     hitsPerPage: hitsPerPage === undefined ? 20 : hitsPerPage, // 20 is the Meilisearch's default limit value. `hitsPerPage` can be changed with `InsantSearch.configure`.
     page: page || 0, // default page is 0 if none is provided
+    finite: !!finite,
```
mdubus, line 19: Why do you use `!!` here?

README.md
```diff
@@ -117,34 +115,18 @@ When placeholder search is set to `false`, no results appears when searching on
...
+
+It requires the usage of the [`Pagination` widget](#-pagination).
 
 Example:
 
 ```js
 { finitePagination: true } // default: false
 ```
 
-⚠️ Meilisearch is not designed for pagination and this can lead to performances issues, so the usage `finitePagination` but also of the pagination widgets are not recommended.<br>
```
brunoocasali, line 146: I think this disclaimer is still accurate... Or at least part of it.

src/utils/number.ts
```diff
@@ -10,3 +10,16 @@ export function ceiledDivision(dividend: number, divisor: number): number {
...
+ * @param  {number} divisor
+ * @returns number
+ */
+export function floorDivision(dividend: number, divisor: number): number {
+  if (divisor > 0) {
+    const NumberPages = Math.floor(dividend / divisor) // total number of pages rounded up to the next largest integer.
+    return NumberPages
+  }
+  return 0
+}
```
brunoocasali, line 25:
```suggestion
export function floorDivision(dividend: number, divisor: number): number {
  if (divisor > 0) {
    return Math.floor(dividend / divisor)
  }

  return 0
}
```

And move this comment `// total number of pages rounded up to the next largest integer.` to the place where you are calling this function, otherwise, name this function with a specialized name.

src/adapter/search-response-adapter/pagination-adapter.ts
```diff
@@ -1,21 +1,55 @@
...
+function adaptNbPages(
+  searchResponse: MeiliSearchResponse<Record<string, any>>,
   hitsPerPage: number
-): Array<Record<string, any>> {
-  if (hitsPerPage < 0) {
-    throw new TypeError(
-      'Value too small for "hitsPerPage" parameter, expected integer between 0 and 9223372036854775807'
-    )
+): number {
+  if (searchResponse.totalPages != null) {
```
brunoocasali, line 12: Some places you opted by `undefined` and others like this `null`, is there any particular reason for that?

src/contexts/pagination-context.ts
bidoubiwa: it forces the conversation to a boolean. So if it is a `falsy`, it becomes `false`

Example: 

`undefined`
! `undefined` => true
!! `undefined` => false

same goes for every falsy https://developer.mozilla.org/en-US/docs/Glossary/Falsy

README.md
bidoubiwa: @gmourier do you think we should keep this?

gmourier: I would remove it to see how the feature is received, we still do something completely different.

- By default, it is limited to the first 1000 results, which should guarantee good performance.
- We removed the hack previously in place, which posed performance problems. According to the testers, the new solution solves this previous pain point.

src/adapter/search-response-adapter/pagination-adapter.ts
bidoubiwa: I dont see a `!= undefined` in my code. Could you show an example?

src/adapter/search-response-adapter/hits-adapter.ts
```diff
@@ -1,24 +1,30 @@
...
   const { primaryKey } = searchContext
-  const { hitsPerPage, page } = paginationContext
-  const paginatedHits = adaptPagination(hits, page, hitsPerPage)
+  const { hits } = searchResponse
+  const {
+    pagination: { finite, page, hitsPerPage },
+  } = searchContext
 
-  let adaptedHits = paginatedHits.map((hit: Record<string, any>) => {
+  if (!finite) {
```
brunoocasali, line 20: Can you help me understand this piece of code, I'm not sure why you need it 👼

src/adapter/search-response-adapter/pagination-adapter.ts
```diff
@@ -1,21 +1,55 @@
...
+  // Finite pagination environment
+  if (page === 0) {
+    // Should not happen but safeguarding in case it does
+    return 0
+  }
+  if (page != null) {
+    return page - 1
+  }
+
+  // Scroll pagination environment
```
brunoocasali, line 34: Can you use the same terms, like `estimated`?

src/adapter/search-response-adapter/total-hits-adapter.ts
```diff
@@ -0,0 +1,20 @@
...
+export function adaptTotalHits(
+  searchResponse: MeiliSearchResponse<Record<string, any>>
+): number {
+  const {
+    hitsPerPage = 0,
+    totalPages = 0,
+    estimatedTotalHits = undefined,
+    totalHits = undefined,
+  } = searchResponse
+  if (estimatedTotalHits !== undefined) {
```
brunoocasali, line 12: Do you think this `estimatedTotalHits` can be null? Because if it is `null` we will have a `null` return.

src/adapter/search-response-adapter/pagination-adapter.ts
bidoubiwa: Absolutely! Thanks for the catch

src/adapter/search-response-adapter/hits-adapter.ts
bidoubiwa:
```
 if (!finite) {
    // In case we are not in a finite environment
    // Hits contains all the documents of every page before the current one.
    const deleteCount = page * hitsPerPage
    // Removes the hits that are not included in the current pagination
    // Example, if we are on page 2, we need to remove the hits from page 1
    hits.splice(0, deleteCount)
    if (hits.length > hitsPerPage) {
      // Since we append the limit of one to know if there is another page
      // We need to remove the last hit in case the length of hits is bigger than the required hitsPerPage
      // Example: limit: 7, hitsPerPage: 6, Hits length: 7. We need to remove the last hit
      hits.splice(hits.length - 1, 1)
    }
  }

```

Is this better?

src/adapter/search-response-adapter/pagination-adapter.ts
bidoubiwa: I don't think `estimated` makes it clear :/ Maybe `Lazy pagination` ?

src/adapter/search-response-adapter/total-hits-adapter.ts
bidoubiwa: As far as all the typed SDK's, it cannot be null

src/adapter/search-response-adapter/pagination-adapter.ts
brunoocasali: I say estimated because of the `estimatedTotalHits` :)

bidoubiwa: We don't use this number in the code base, it does help the user understand for me

PR #896, opened 2022-12-01
Keep both _geo and _geoloc fields in hits and hits._formatted

bidoubiwa:
Instantsearch geo widget requires the geo field to be named `_geoloc`. Nonetheless, the field returned from meilisearch is `_geo`. 

Previously we replaced `_geo` with `_geoloc` at the root of the hits (so not in `_formatted`) which in itself was a bug.

Since `instantsearch` needs `_geoloc` but we also want to avoid confusing the user to why their `_geo` field disappeared, we decided to keep both `_geo` and `_geoloc` files at the root of the hit but also in _formatted (of course only if there was a `_geo` field present to begin with).


mdubus (approved):
LGTM ✨🦕

bidoubiwa:
bors merge

PR #916, opened 2022-12-08
Multi-index search: fix placeholdersearch not working properly with filters and pagination

bidoubiwa:
`placeholderSearch` lets you enable (default) or disable the rendering of the hits when an empty search is made.
An empty search is a search without any query and without any filters.

When `placeholderSearch` was set to false, hits did not render when a filter was required.
When `placeholderSearch` was set to false, the pagination was rendered of is placeholder search was set to true.

With this PR:

- When placeholderSearch is set to false, If you add a filter, by clicking on a facet for example, hits are rendered
- When placeholderSearch is set to false, when the query and the filters are empty, the pagination does not render any page. 
    - In finitePagination this is what is rendered:
        Screenshot 2022-12-08 at 15 54 54
   - in scroll pagination: the load more button is disabled
        Screenshot 2022-12-08 at 15 55 51

### DIFF
--- a/src/adapter/search-request-adapter/search-params-adapter.ts
+++ b/src/adapter/search-request-adapter/search-params-adapter.ts
@@ -1,4 +1,9 @@
-import type { MeiliSearchParams, SearchContext } from '../../types'
+import type {
+  MeiliSearchParams,
+  SearchContext,
+  Filter,
+  PaginationState,
+} from '../../types'
 
 import {
   adaptGeoPointsRules,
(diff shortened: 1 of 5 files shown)


src/adapter/search-request-adapter/search-resolver.ts
```diff
@@ -54,13 +52,9 @@ export function SearchResolver(
         )
       }
 
-      // query can be: empty string, undefined or null
-      // all of them are falsy's
-      if (!placeholderSearch && !query) {
-        searchResponse.hits = []
-      }
```
bidoubiwa, line 61: Should not be handled here. Moved to the query params handler

src/types/types.ts
```diff
@@ -78,11 +78,11 @@ export type SearchContext = Omit<InstantSearchParams, 'insideBoundingBox'> &
     defaultFacetDistribution: FacetDistribution
     pagination: PaginationState
     indexUid: string
+    placeholderSearch: boolean
+    keepZeroFacets: boolean
     insideBoundingBox?: InsideBoundingBox
-    keepZeroFacets?: boolean
     cropMarker?: string
     sort?: string
-    placeholderSearch?: boolean
```
bidoubiwa, line 85: In the code, they can never be undefined. Updating this type to reflect that

src/adapter/search-request-adapter/search-params-adapter.ts
```diff
@@ -1,18 +1,36 @@
...
   query?: string,
   placeholderSearch?: boolean
+): boolean {
+  if (!placeholderSearch && !query) {
+    if (!filter || filter.length === 0) {
+      return false
+    }
+  }
+  return true
+}
```
mdubus: What happens if `if (!placeholderSearch && !query)` is true, but `if (!filter || filter.length === 0) {` is false?

bidoubiwa: then the pagination is required, both must be true to disable pagination

brunoocasali (approved):

bidoubiwa:
bors merge

Issue #937, opened 2023-01-10
Migrating to a turborepo architecture


bidoubiwa:
## What is turborepo

Turborepo is a high-performance build system for JavaScript and TypeScript codebases. 
It manages and publishes multiple JavaScript/TypeScript packages from the same repository (mono repo).
Examples: [supabase](https://github.com/supabase/supabase)

## Why do we need turbo

Currently, this repository contains only one package: `instant-meilisearch`. Instant-meilisearch is a search client adapting the search requests from different tools, currently only [instantsearch.js](https://github.com/algolia/instantsearch), to Meilisearch compatible search requests. Same for search responses.

Our ambition is to be able to also add other adapters. For example, to be able to use meilisearch with [autocomplete](https://github.com/algolia/autocomplete/tree/next).

In order to ease the development of these adapters, having them all in the same repo helps a lot. In this case, we can ensure that the adapter of `autocomplete` works well with the `instantMeilisearch` client without having to wait on publishing or creating symlinks between repos.

## How does it impact our current structure

Our current architecture is not compliant to a turborepo architecture. This means that a lot of files are going to be moved around and configuration files will need to be added and updated.

Nonetheless, this should not impact at all our end user. 

## What needs to be done
- [ ] Change the name of the repo
- [x] Move instant-meilisearch inside a `package` folder
   - [x] source code
   - [x] tests
   - [x] readme
   - [x] package.json
   - [x] tsconfig
- [ ] Playgrounds:
   - [ ] Ensure hot reloading still works
     - [x] javascript
     - [x] geosearch
     - [x] react
     - [ ] angular
     - [ ] vue
   - [x] Ensure linting works
      - [x] javascript
      - [x] geosearch
      - [x] react
      - [x] angular
      - [x] vue
   - [x] Ensure build works
      - [x] javascript
      - [x] geosearch
      - [x] react
      - [x] angular
      - [x] vue
- [x] Tests:
   - [x] Tests must be added in the instant-meilisearch package
- [ ] Update the contributing guide
- [ ] Create a new readme containing all the plugins
- [x] Cleanup the package.json at the root
- [x] Cleanup the package.json of instant-melisearch
- [x] update tsconfig
- [x] update eslint
- [x] Add turborepo config
- [x] Update CI's
- [x] Update Publish CI
- [x] Update dependabot CI to automatically accept package updates in the playgrounds

PR #940, opened 2023-01-10
Move instant-meilisearch to packages folder

bidoubiwa:
Changes in this PR:

To be compliant with the workspace architecture required by turborepo. This PR moves instant-meilisearch inside its own package:

Sources, tests, README, tsconfig, package.json files are moved to `/packaes/instant-meilisearch`


brunoocasali:
Thinking about the structure naming: `instant-meilisearch/packages/instant-meilisearch` it would be repetitive, maybe you can use a different naming? Or do you prefer to change the repository name instead?

bidoubiwa:
> Thinking about the structure naming: instant-meilisearch/packages/instant-meilisearch it would be repetitive, maybe you can use a different naming? Or do you prefer to change the repository name instead?

The repo's name should definitely be changed to something like `meilisearch-js-plugins`.

I made a list here of the things that needs to be done: https://github.com/meilisearch/instant-meilisearch/issues/937

brunoocasali (approved):
🔥🔥

PR #941, opened 2023-01-12
Move instant meilisearch to packages

bidoubiwa:
Add the configuration files for turborepo initialization.

- Create a root `package.json` that is named `root` is private and contains the workspaces.
- Add a `turbo.json` configuration file with the tasks we want to be able to run in the multiple packages/app (see [doc](https://turbo.build/repo/docs/core-concepts/monorepos/running-tasks#turborepo-can-multitask)).
-


bidoubiwa:
duplicated pr #940

PR #942, opened 2023-01-12
Add turbo configuration

bidoubiwa:
Add the configuration files for turborepo initialization.

- Create a root `package.json` that is named `root` is private and contains the workspaces.
- Add a `turbo.json` configuration file with the tasks we want to be able to run in the multiple packages/app (see [doc](https://turbo.build/repo/docs/core-concepts/monorepos/running-tasks#turborepo-can-multitask)).
-


package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "root",
```
brunoocasali, line 2: 👀

brunoocasali (approved):
It seems very simple 🙈

package.json
bidoubiwa: It's the convention :')

PR #943, opened 2023-01-12
Update playgrounds to turbo

bidoubiwa:
Update playgrounds in order to be able to run them in hot loading with the packages:

- Updated: react, angular, vanilla and geosearch
- Removed: HTML as it was redundant with vanilla 
- Removed: vue as it was overcomplicated, it'll be added in a future PR


playgrounds/geo-javascript/src/app.js
```diff
@@ -1,4 +1,4 @@
-import { instantMeiliSearch } from '../../../src/client/index'
+import { instantMeiliSearch } from '@meilisearch/instant-meilisearch'
```
brunoocasali, line 1: This will use the local version and not the npm one, right?

bidoubiwa: Yes! Local version :)

bidoubiwa: https://turbo.build/repo/docs/handbook/workspaces#workspaces-which-depend-on-each-other

brunoocasali (approved):
👍

PR #945, opened 2023-01-12
Enable linting

bidoubiwa:
This PR creates the correct configuration to run the linting in all projects. 

In this PR, I created a package named `eslint-config-meilisearch`, in which the linting rules are present. This dependency is then used in every package/app.

To better understand the process, I suggest reading [this documentation](https://turbo.build/repo/docs/handbook/linting/eslint).

It would be nice in the future to move this eslint config in its own repo and use it in every JS repository we have. Similarely to what [algolia has done](https://github.com/algolia/eslint-config-algolia). After which, we would be able to remove a ton of eslint dependency on all these repo's and ensure we have the same code styling on all of them. Anyway, I'll create an issue with this suggestion in the integration-guide.

- ⚠️ `/env/react` is removed from the directories that are being lint on purpose.  This repository should be moved to `/playgrounds` but that will be done in another PR.


playgrounds/vue3-ts/src/App.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div>
     <header class="header">
-      <h1 class="header-title">PLOUF MeiliSearch + Vue 3 InstantSearch</h1>
+      <h1 class="header-title">MeiliSearch + Vue 3 InstantSearch</h1>
```
bidoubiwa, line 4: Not related to the PR 🙈

packages/instant-meilisearch/package.json
```diff
@@ -104,6 +82,7 @@
     "shx": "^0.3.3",
     "ts-jest": "^27.0.7",
     "tslib": "^2.4.1",
-    "typescript": "^4.9.3"
+    "typescript": "^4.9.3",
+    "eslint-config-meilisearch": "*"
```
brunoocasali, line 86: Why is it not `@meilisearch/eslint-config`? 😄

playgrounds/geo-javascript/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
brunoocasali, line 1: Can this be shared somehow?

playgrounds/vue3-ts/src/App.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div>
     <header class="header">
-      <h1 class="header-title">PLOUF MeiliSearch + Vue 3 InstantSearch</h1>
```
brunoocasali, line 4: 👀 Ha! Spotted!

packages/eslint-config-meilisearch/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "eslint-plugin-jsx-a11y": "^6.4.1",
+    "eslint-plugin-node": "^11.1.0",
+    "eslint-plugin-prettier": "^3.3.1",
+    "eslint-plugin-promise": "^4.3.1",
+    "eslint-plugin-react": "^7.22.0",
+    "eslint-plugin-standard": "^5.0.0",
+    "eslint-plugin-vue": "^7.7.0",
+    "typescript": "^4.7.4"
+  }
+}
```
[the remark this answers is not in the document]
bidoubiwa: I did not updated the dependabot configuration yet, it's in the todo. https://github.com/meilisearch/instant-meilisearch/issues/937

But I would like that!

packages/instant-meilisearch/package.json
bidoubiwa: It's because of the eslint convention naming. If tomorrow I want to publish this configuration to npm it needs to follow that naming convention

playgrounds/geo-javascript/.eslintrc.js
bidoubiwa: What part?

playgrounds/vue3-ts/src/App.vue
bidoubiwa: 😂🙈

packages/eslint-config-meilisearch/package.json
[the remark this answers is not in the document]
brunoocasali: As you can see I'm highly interested on that ahaha :)

playgrounds/geo-javascript/.eslintrc.js
brunoocasali: The whole file, between the other playgrounds.

brunoocasali:
I appreciate this idea **It would be nice in the future to move this eslint config in its own repo and use it in every JS repository we have** a lot. In the integrations team I'm sure I now we can do that, but what about the cloud, it is possible to share the same rules?

playgrounds/geo-javascript/.eslintrc.js
bidoubiwa: No they should not rely on out-of-their scope files. They are already using the configuration from `/eslint-config-meilisearch` but they need `.eslintrc` to trigger it in their scope. 

See the structure in this doc: https://turbo.build/repo/docs/handbook/linting/eslint

bidoubiwa:
> I appreciate this idea It would be nice in the future to move this eslint config in its own repo and use it in every JS repository we have a lot. In the integrations team I'm sure I now we can do that, but what about the cloud, it is possible to share the same rules?

Yes I suggested that in the body of the issue! We could absolutely sync that with the cloud as well if they like my rules, or discuss with them about it.

brunoocasali (approved):

packages/eslint-config-meilisearch/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "eslint-config-meilisearch",
+  "version": "0.0.0",
```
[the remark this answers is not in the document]
bidoubiwa: Since we are not publishing it, its not needed. See example from turbo's directory: https://github.com/vercel/turbo/blob/main/examples/basic/packages/eslint-config-custom/package.json

PR #949, opened 2023-01-17
Move local react playground to playground dir

bidoubiwa:
There was a `react` playground nested in the `__tests__` sub-folder of instant-meilisearch. Nonetheless, with the turbo architecture it is best to move that playground to the playground directory. 

Further changes will be done in another PR to avoid losing the history of the files.


mdubus (approved):
LGTM ✨🦕

PR #950, opened 2023-01-17
Update local react playground 

bidoubiwa:
The previously moved local react playground #949 needed to be updated to be working in the turbo architecture.


brunoocasali (approved):
🐪 🍣 LGTM!

mdubus (approved):
LGTM ✨🦕

PR #951, opened 2023-01-18
Fix building fail in @meilisearch/vue-playground

bidoubiwa:
the `vue` playground could not be build because of missing type files not related to instant-meilisearch. This PR adds a work-arround to ignore the issue.


alallema (approved):
🙈

PR #954, opened 2023-01-19
Fix end 2 end tests for turborepo migration

bidoubiwa:
Since we moved the web app that was used for the end 2 end tests to another directory (#949) , we need to update the way we trigger them.


playgrounds/local-react/package.json
```diff
@@ -8,7 +8,9 @@
     "dev": "parcel serve index.html --open --port 9999",
     "build": "parcel build index.html",
     "setup": "node setup.js",
-    "lint": "eslint --ext .jsx,.js ."
+    "lint": "eslint --ext .jsx,.js .",
+    "test:e2e": "concurrently --kill-others -s first \"yarn dev\" \"cypress run --env playground=local\"",
```
brunoocasali, line 12: Where do you use the `playground=local` env var?

brunoocasali: And the commit could have a description like this:

```
Fix trigger command of e2e tests after monorepo migration
```

brunoocasali (approved):

PR #955, opened 2023-01-19
Fix typing tests for turborepo migration

bidoubiwa:
Types tests were failing on @meilisearch/instant-meilisearch because of dependencies from the angular playground. These have been removed and types test were added on every typescript package (vue and angular)


bidoubiwa:
bors merge

PR #957, opened 2023-01-25
Update the way we trigger tests in CI's

bidoubiwa:
The way we trigger tests were changed, see #953. While they were updated in the `tests.yml` CI file, they also needed to be updated in the pre-release and beta CI's


brunoocasali (approved):
Random opinion: About having a playground as a "test", to me it would be good if we have two things "one playground" and the other can be a react test environment...

Issue #959, opened 2023-01-25
Use changeset to handle publishing of packages


bidoubiwa:
With the migration to #939 we need to change the way we handle publishes on github but also on `npm`. 

The tool [suggested by turborepo](https://turbo.build/repo/docs/handbook/publishing-packages/versioning-and-publishing) is changeset. 

Changeset automates a lot of the tedious actions, for example updating the version manually or creating relevant information on what the PR does.

## Changeset 
Changing to changeset requires the contributor to add a [`changeset`](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md) in its PR. A changeset A changeset is a piece of information about changes made in a branch or commit. It generates a unique file containing different information.

For example: 
```
---
"@meilisearch/instant-meilisearch": minor
---

I did something in this PR
```

## Version Update

When running `changeset version` all the files are merged into the CHANGELOG of the related package. In the case above it updates (or creates) the CHANGELOG of the `@meilisearch/instant-meilisearch` package (and not the others).

In our case it would generate this:
```
# @meilisearch/instant-meilisearch

## 0.11.0

### Minor Changes

- I did something

```

It also updates the version in the related package based on whether the changesets asked for a major, minor or patch update. In our case, we asked for a `minor` update, so the package.json of `@meilisearch/instant-meilisearch` has its version updated to `0.11.0`.

## Publish

Finally, based on the packages that saw their version updated, the publish function will only publish the affected packages.

## Process

We want users to not use `changeset publish` or `changeset version` locally. This should be handled by the CI. 
`changesets` has a [dedicated github action](https://github.com/changesets/action)to handle that.

In this action we want the following: 
- On each push to main, the CI should update a PR it created with the updated changelogs and versions. [Example](https://github.com/emotion-js/emotion/pull/2885)
- When the PR is merged, the CI should generate both a github release for the related packages ([see example](https://github.com/emotion-js/emotion/releases)) and publish then on `npm`. 

## Todo
- [ ] Add `changesets` configuration to the repo
- [ ] Update the contributing guide with the new required changeset that has to be added on every PR
- [ ] Update the publish CI to follow the process described above
- [ ] Handle pre-releases and beta versions

bidoubiwa:
fixed by #965

PR #960, opened 2023-01-25
Add changesets configuration to handle versioning and publishing

bidoubiwa:
See #959


.github/workflows/publish.yml
```diff
@@ -1,32 +1,32 @@
...
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
           node-version: 16
           registry-url: https://registry.npmjs.org/
-      - name: Check release validity
-        run: sh .github/scripts/check-release.sh
-      - name: Check tag format
-        run: sh .github/scripts/check-tag-format.sh "${{ github.event.release.prerelease }}"
```
brunoocasali, line 18: No need to check the tag format anymore?

package.json
```diff
@@ -31,6 +33,8 @@
     "turbo": "^1.7.0"
   },
   "dependencies": {
+    "@changesets/cli": "^2.26.0",
     "instantsearch.css": "^8.0.0"
-  }
+  },
+  "version": "0.0.0"
```
brunoocasali, line 39: Can you move to the top? 😄 Also, 0.0.1 O think is more appropriate (even if is going to be eternal) :)

packages/instant-meilisearch/scripts/update_version.js
```diff
@@ -0,0 +1,13 @@
+const pkg = require('../package.json')
+const fs = require('fs')
+const path = require('path')
+
+const version = pkg.version
+
```
brunoocasali, line 6:
```suggestion
```

packages/instant-meilisearch/scripts/update_version.js
```diff
@@ -0,0 +1,13 @@
...
+const pckVersion = `export const PACKAGE_VERSION = '${version}'
+`
+
+const p = path.resolve(__dirname, '../src/package-version.ts')
+
+console.log(p)
+fs.writeFileSync(p, pckVersion)
```
brunoocasali, line 13:
```suggestion
const fileContents = `export const PACKAGE_VERSION = '${version}'
`

const filePath = path.resolve(__dirname, '../src/package-version.ts')

console.log(filePath)
fs.writeFileSync(filePath, fileContents)
```

brunoocasali (approved):
Just some recommendations :)

package.json
bidoubiwa: I added `0.0.0` as this is what seems to be the convention in the doc and the repo examples I came across.
- https://turbo.build/repo/docs/reference/codemods#add-package-manager

.github/workflows/publish.yml
bidoubiwa: I'll maybe add it in the PR that tries the different beta types if it is relevant

PR #987, opened 2023-02-08
Exit pre-release mode for next release

bidoubiwa:
Since #939 has been released in pre-release mode and then pushed to main, we now need to exit pre-release mode.

As per [the documentation](https://github.com/changesets/changesets/blob/main/docs/prereleases.md)


brunoocasali (approved):

bidoubiwa:
bors merge

Issue #1013, opened 2023-02-27
Exception thrown when sorting results (v0.11.0)


Strift:
**Description**

Sorting results using the `AisSortBy` Vue component does not work since v0.11.0. The same code used to work in v0.10.0.

**Expected behavior**

Results to be sorted.

**Current behavior**

The following errors are thrown:

```
instant-meilisearch.esm.js:971 TypeError: Cannot read properties of undefined (reading 'category')
    at fillMissingFacetValues (instant-meilisearch.esm.js:743:32)
    at adaptFacetDistribution (instant-meilisearch.esm.js:762:16)
    at adaptSearchResponse (instant-meilisearch.esm.js:782:29)
    at Proxy.<anonymous> (instant-meilisearch.esm.js:962:53)
    at step (instant-meilisearch.esm.js:73:23)
    at Object.next (instant-meilisearch.esm.js:54:53)
    at fulfilled (instant-meilisearch.esm.js:44:58)

instant-meilisearch.esm.js:972 Uncaught (in promise) Error: TypeError: Cannot read properties of undefined (reading 'category')
    at Proxy.<anonymous> (instant-meilisearch.esm.js:972:35)
    at step (instant-meilisearch.esm.js:73:23)
    at Object.next (instant-meilisearch.esm.js:54:53)
    at fulfilled (instant-meilisearch.esm.js:44:58)
```

**Screenshots or Logs**

Code is available on this branch: https://github.com/meilisearch/ecommerce-demo/tree/1-install-meilisearch-module

**Environment (please complete the following information):**

- OS: Mac OS
- Browser: Brave
- Meilisearch version: v1.0.0
- instant-meilisearch version: v.0.11.0
- vue-instantsearch version: v4.8.5

PR #1016, opened 2023-03-01
Fix initial facet distribution when using sortby

bidoubiwa:
Fixes: #1013 

Additionally I removed a file that was not used in the code-base and is basically an old version of `init-facets-distribution`


packages/instant-meilisearch/src/client/instant-meilisearch-client.ts
```diff
@@ -98,7 +98,7 @@ export function instantMeiliSearch(
           const adaptedSearchResponse = adaptSearchResponse<T>(
             searchResponse,
             searchContext,
-            initialFacetDistribution[searchRequest.indexName]
+            initialFacetDistribution[searchContext.indexUid]
```
brunoocasali, line 101: No test?! 👀

Strift: From my modest understanding of what's happening, it looks like the correct thing to do to retrieve the initial facet distribution.

Strift (commented):
The solution looks good to me — but I'm unsure (and lacking knowledge to understand) what's the impact of the removed file.

packages/instant-meilisearch/src/client/instant-meilisearch-client.ts
bidoubiwa: Done!

bidoubiwa:
bors merge

PR #1039, opened 2023-03-21
Fix pre-release CI not triggering

bidoubiwa:
When a push occurs on a branch pointing to a bump branch, the `push` action does not trigger as it does not match the required branch patterns. When a pull_request activity on a bump branch occurs, the pre-release tests were triggered but because of the wrong if condition, the jobs were not executed.

The if condition should ensure that the base branch corresponds to the pull_request branch patterns.

Which is now the case


alallema (approved):
LGTM! ✨

bidoubiwa:
bors merge

PR #1048, opened 2023-03-29
Use _geoBoundingBox when creating the geo search filter

bidoubiwa:
As per the specification: https://github.com/meilisearch/specifications/pull/223

We used to transform the `insideBoundingBox` provided by `instantsearch` to a filter using `_geoRadius`. This required a lot of maths and was not the behavior expected by the user.

With the introduction of the `_geoBoundingBox` filter in Meilisearch, we can now use that instead of `_geoRadius`. 

⚠️ Search results may differ as the geo search algorithm changed


alallema (approved):
LGTM!
Just one question

packages/instant-meilisearch/src/adapter/search-request-adapter/geo-filter-adapter.ts
```diff
@@ -0,0 +1,49 @@
...
+  aroundRadius,
+  minimumAroundRadius,
+}: InstantSearchGeoParams): string | undefined {
+  let middlePoint: string[] | undefined
+  let radius: number | undefined
+  let filter: string | undefined
+
+  if (aroundLatLng) {
+    const [lat3Str, lng3Str] = aroundLatLng.split(',')
+    const lat3 = Number.parseFloat(lat3Str).toFixed(5)
```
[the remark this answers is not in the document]
bidoubiwa: Yes, it's [in the standard library](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed)

alallema (approved):

packages/instant-meilisearch/src/adapter/search-request-adapter/__tests__/geo-rules.test.ts
```diff
@@ -1,82 +1,82 @@
-import { adaptGeoPointsRules } from '../geo-rules-adapter'
+import { adaptGeoFilter } from '../geo-rules-adapter'
 
-test('Adapt geoPoints rules without a boundingBox', () => {
-  const rules = adaptGeoPointsRules()
-  expect(rules).toBeUndefined()
+test('Adapt geoPoints filter without a boundingBox', () => {
```
mdubus, line 3: `geoPoints filter` or `geo filter` (as in `adaptGeoFilter`)? I'm confused now :D

packages/instant-meilisearch/src/adapter/search-request-adapter/search-params-adapter.ts
```diff
@@ -176,15 +173,26 @@ export function MeiliParamsCreator(searchContext: SearchContext) {
         meiliSearchParams.sort = [sort]
       }
     },
-    addGeoSearchRules() {
-      const geoSearchContext = createGeoSearchContext(searchContext)
-      const geoRules = adaptGeoPointsRules(geoSearchContext)
+    addGeoSearchFilter() {
```
mdubus, line 176: Same here, I'm confused with the naming. 
Upper you do the following:
`import { adaptGeoFilter } from './geo-rules-adapter'`
And here you're writing `addGeoSearchFilter`. In the code you sometimes have `geoXXX`, `geoPointsXXX` and `geoSearchXXX`. I feel like the `geoXXX` is no precise enough and we don't really know what it is exactly. 
Is this intended or is something missing in the naming?

packages/instant-meilisearch/src/adapter/search-request-adapter/__tests__/geo-rules.test.ts
bidoubiwa: I agree, let me rename all the tests

mdubus (approved):
LGTM ✨🦕

PR #1049, opened 2023-03-29
Remove unecessary types

bidoubiwa:
As per the last update of `meilisearch-js`, new types were introduced that removes the need of instant-meilisearch re-creating them.


alallema (approved):
LGTM!

packages/instant-meilisearch/src/adapter/search-response-adapter/pagination-adapter.ts
```diff
@@ -1,11 +1,11 @@
 import type {
-  MeiliSearchResponse,
+  MultiSearchResult,
   InstantSearchPagination,
   PaginationState,
 } from '../../types'
 
 function adaptNbPages(
-  searchResponse: MeiliSearchResponse<Record<string, any>>,
+  searchResponse: MultiSearchResult<Record<string, any>>,
```
mdubus, line 8: I find it weird reading `searchResponse: MultiSearchResult`. Are we receiving a response? A result? And for single or multi search?
Would it be worth it renaming `searchResponse` into `multiSearchResult`?
Feel free to ignore my comment if it's irrelevant :)

packages/instant-meilisearch/src/types/types.ts
```diff
@@ -2,24 +2,19 @@ import type { SearchClient } from 'instantsearch.js'
...
-}
+export type { AlgoliaMultipleQueriesQuery, MultiSearchResult }
 export type { SearchResponse as AlgoliaSearchResponse } from '@algolia/client-search'
 
 export type {
   Filter,
   FacetDistribution,
   MeiliSearch,
   FacetStats as MeiliFacetStats,
+  MultiSearchQuery as MeiliSearchMultiSearchParams,
```
mdubus, line 17: Same here, it seems weird 🤔 
Upper you're importing the following:
`MultiSearchQuery as MeiliSearchMultiSearchParams,`
So can't you just import and export `MultiSearchQuery` ?

packages/instant-meilisearch/src/adapter/search-response-adapter/pagination-adapter.ts
bidoubiwa: Indeed! I was avoiding doing in this PR to avoid putting to many out of context changes. But it should definitely be changed

packages/instant-meilisearch/src/types/types.ts
bidoubiwa: Same as above, it's to avoid renaming variables in an unrelated PR. But I definitely agree with you

mdubus (approved):
LGTM ✨🦕

PR #1092, opened 2023-05-02
Change the cmd to run the script in the node-playground

bidoubiwa:
The build command in the node playground was run whenever the release was triggered. But in reality the node playground does not need to be built, it rather runs a script to see if it can interact correctly with instant meilisearch.

This PR changes the command from build to dev so that turbo build does not trigger it. It also adds a line in the test ci to ensure the script runs correctly

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -99,6 +99,8 @@ jobs:
         run: yarn test
       - name: Build all the playgrounds and the packages
         run: yarn build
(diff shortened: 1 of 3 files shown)


CaroFG (approved):

bidoubiwa:
bors merge

CaroFG (approved):

bidoubiwa:
bors merge

Issue #1083, opened 2023-04-26
Filtering using a key which has spaces (e.g. "In Stock") raises an error


rub1e:
**Description**
Filtering using a key which has spaces (e.g. "Lens Color") raises an error

**Expected behavior**
You should be able to filter using keys with spaces

**Current behavior**
MeilisearchApiError is returned

**Screenshots or Logs**
```
[MeiliSearchApiError]: Inside `.queries[0]`: Was expecting an operation `=`, `!=`, `>=`, `>`, `<=`, `<`, `IN`, `NOT IN`, `TO`, `EXISTS`, `NOT EXISTS`, `_geoRadius`, or `_geoBoundingBox` at `Lens Color=\"Red\"`.
1:17 Lens Color="Red"
    at /home/rubie/oco-hydrogen/node_modules/@meilisearch/instant-meilisearch/dist/instant-meilisearch.umd.js:359:25
    at step (/home/rubie/oco-hydrogen/node_modules/@meilisearch/instant-meilisearch/dist/instant-meilisearch.umd.js:266:25)
    at Object.next (/home/rubie/oco-hydrogen/node_modules/@meilisearch/instant-meilisearch/dist/instant-meilisearch.umd.js:215:22)
    at fulfilled (/home/rubie/oco-hydrogen/node_modules/@meilisearch/instant-meilisearch/dist/instant-meilisearch.umd.js:174:32)
    at processTicksAndRejections (node:internal/process/task_queues:95:5) {
  code: 'invalid_search_filter',
  type: 'invalid_request',
  link: 'https://docs.meilisearch.com/errors#invalid_search_filter',
  httpStatus: 400
}
```

**Environment (please complete the following information):**
- OS: Ubuntu
- Browser: All
- Meilisearch version: 1.1
- instant-meilisearch version: 0.12.0
- instantsearch.js version: 4.51.1

[Edited with better example]

oluademola:
Hello @rub1e. Sorry for the delay. I think this is due to a badly formatted filter, if you use quotes around the key or attribute it should work. For example: `"filter": "'in stock' = true"`

rub1e:
Hi @oluademola, thanks for getting back to me. Apologies for lack of clarity, I can now see what's causing the error:

The error occurs when using the instantsearch widgets like ToggleRefinement or RefinementList.

So this causes the error

`<ToggleRefinement attribute="In Stock" />`
`<RefinementList attribute="Lens Color" />`

[...]

PR #1105, opened 2023-05-17
Fix multi words filters not formatted correctly

bidoubiwa:
fixes: #1083 

Facets containing multiple words, for example `In stock`, were wrongly formatted. Meilisearch requires multi-words in filters to be wrapped in double quotes. The solution provided is to wrap every facet attribute and facet value in double quotes.

Before:
'In stock="value"'
After
'"In stock"="value"'

### DIFF
--- a/packages/instant-meilisearch/src/adapter/search-request-adapter/__tests__/search-params.test.ts
+++ b/packages/instant-meilisearch/src/adapter/search-request-adapter/__tests__/search-params.test.ts
@@ -26,12 +26,23 @@ describe('Parameters adapter', () => {
     })
 
     expect(searchParams.filter).toStrictEqual([
-      ['genres="Drama"', 'genres="Thriller"'],
-      ['title="Ariel"'],
+      ['"genres"="Drama"', '"genres"="Thriller"'],
+      ['"title"="Ariel"'],
     ])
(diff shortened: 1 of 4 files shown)


packages/instant-meilisearch/src/adapter/search-request-adapter/__tests__/search-params.test.ts
```diff
@@ -26,14 +26,25 @@ describe('Parameters adapter', () => {
...
   })
 
+  test('adapting multi-word filters', () => {
+    const searchParams = adaptSearchParams({
+      ...DEFAULT_CONTEXT,
+      facetFilters: [['My Genre:Science Fiction']],
+    })
+
+    expect(searchParams.filter).toStrictEqual([
+      ['"My Genre"="Science Fiction"'],
```
brunoocasali, line 44: I didn't know that works, cool!

brunoocasali (approved):
🦖 approves! LGTM!

packages/instant-meilisearch/src/adapter/search-request-adapter/__tests__/search-params.test.ts
bidoubiwa: me neither haha

bidoubiwa:
bors merge

PR #1141, opened 2023-06-06
Add missing version script for autocomplete client

bidoubiwa:
The version script required by the changeset CI to update the versions of the package was missing.


packages/autocomplete-client/scripts/update_version.js
```diff
@@ -0,0 +1,12 @@
...
+const fs = require('fs')
+const path = require('path')
+
+const version = pkg.version
+const fileContents = `export const PACKAGE_VERSION = '${version}'
+`
+
+const filePath = path.resolve(__dirname, '../src/package-version.ts')
+
+console.log(filePath)
```
alallema, line 11: If it's a debug log, maybe a sentence will be great. If not it should be remove

packages/autocomplete-client/scripts/build.js
```diff
@@ -0,0 +1,65 @@
...
+// Definitions by: ${author} <https://github.com/${githubUserUrl}>
+// Definitions: ${repoUrl}
+// TypeScript Version: ${tsVersion}
+`.replace(/^\s+/gm, '')
+}
+
+/**
+ *
+ * @param {string} path
+ * @param {string | Blob} data
```
alallema, line 54: What is `Blob`?

alallema (commented):
I don't understand the purpose of this script. It said to build and update the version for the second, but it seems to write some files.
Sorry, js isn't really my strong suit.

packages/autocomplete-client/scripts/update_version.js
bidoubiwa: I'm removing it

packages/autocomplete-client/scripts/build.js
bidoubiwa: Here you go: https://developer.mozilla.org/en-US/docs/Web/API/Blob

alallema (approved):
LGTM!

PR #1144, opened 2023-06-06
Prototype beta/prototype search for facet values

bidoubiwa:
## New

Add compatibility with the `searchable` parameter of the [`RefinementList`](https://www.algolia.com/doc/api-reference/widgets/refinement-list/js/) widget

## TODO

- [ ] work keepzerofacets compatibility


gmourier:
I was testing the search for facet values prototype with the instant-meilisearch refinement list.

So far, it works well. The only feedback I can think of is that the search box stays filled with my facet query when documents change because of further action (like an update of the search bar or a click on another facet value)

In that case, I would like the RefinementList search box to be emptied

PR #1182, opened 2023-06-27
Put back instant-meilisearch templates in its package

bidoubiwa:
During the migration to turbo-repo the `templates` where not moved in the instant-meilisearch package.
This leads to the templates not being available from the npm package.


brunoocasali (approved):
LGMT

bidoubiwa:
bors merge

PR #1242, opened 2023-08-30
Update dependabot changing ignore list to allow list

bidoubiwa:
To avoid being spammed with none-critical dependencies updates suggested by dependabot, we are changing from `ignore` to `allow` in the dependabot file. Ensuring we only update the critical dependencies.

```
- dependency-name: meilisearch
```

meilisearch is kept as it is the only one that is builded with the package. Updates of this library ensure that the library includes all new features provided by meilisearch-js.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,8 +5,8 @@ updates:
     schedule:
       interval: "monthly"
     labels:
-      - 'skip-changelog'
-      - 'dependencies'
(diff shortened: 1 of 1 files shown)


alallema (commented):
LGTM! 🚀

alallema (approved):

brunoocasali:
bors merge

PR #1251, opened 2023-09-25
Update dependabot.yml

curquiza:
We don't want to skip this information in the changelog anymore since only Meilisearch version can be updated


bidoubiwa (approved):

curquiza:
bors merge

PR #1259, opened 2023-10-04
Escape filter backslashes and quotes

### LINKED ISSUE #1256: Error when choosing a facet containing quotation marks (`"` character)
**Description**
When using a facet that contains `"` characters, the `filter` parameter has it's backslashes not properly escaped (I think), which causes an error.

<details>
<summary>Query body</summary>

```
{
  "queries": [
    {
      "q": "",
      "indexUid": "i",
      "facets": [
        "document_type",
        "imprint.publication_place",
        "imprint.publisher_name",
        "languages",
        "library",
        "other_authors",
        "primary_author",
        "publication_year",
        "religion",
        "subjects"
      ],
      "attributesToHighlight": [
        "*"
      ],
      "highlightPreTag": "__ais-highlight__",
      "highlightPostTag": "__/ais-highlight__",
      "hitsPerPage": 10,
      "page": 1,
      "filter": [
        [
          "\"imprint.publisher_name\"=\"\"MAgyar Őskutatás\" K.\""
        ]
      ]
    },
    {
      "q": "",
      "indexUid": "i",
      "facets": [
        "imprint.publisher_name"
      ],
      "attributesToHighlight": [
        "*"
      ],
      "highlightPreTag": "__ais-highlight__",
      "highlightPostTag": "__/ais-highlight__",
      "hitsPerPage": 0,
      "page": 1
    }
  ]
}
```
</details>

<details>
<summary>Error message</summary>

```
MeiliSearchApiError: Inside `.queries[0]`: Found unexpected characters at the end of the filter: `MAgyar Őskutatás\" K.\"`. You probably forgot an `OR` or an `AND` rule.
28:49 "imprint.publisher_name"=""MAgyar Őskutatás" K."
```
</details>

**Environment:**
- Meilisearch version: v1.4.0, v1.3.3, v1.3.5
- instant-meilisearch version: v0.13.6
- instantsearch.js version: v4.57.0

flevi29:
# Pull Request

## Related issue
Fixes #1256

## What does this PR do?
- Fixes issue where backslashes ("\\") and quotes (`"`) are not escaped in filters
- Also numeric filter attributes are now always wrapped in quotes

## PR checklist
- [x] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [x] Have you read the contributing guidelines?
- [x] Have you made sure that the title is accurate and descriptive of the changes?

### DIFF
--- a/packages/instant-meilisearch/__tests__/filter.test.ts
+++ b/packages/instant-meilisearch/__tests__/filter.test.ts
@@ -11,43 +11,16 @@ describe('Instant Meilisearch Browser test', () => {
     await meilisearchClient.waitForTask(deleteTask.taskUid)
     await meilisearchClient
       .index('movies')
-      .updateFilterableAttributes(['genres', 'title'])
+      .updateFilterableAttributes([
+        'genres',
+        'title',
+        'release_date',
+        'escape_\\_"me"',
+      ])
     const documentsTask = await meilisearchClient
       .index('movies')
       .addDocuments(dataset)
     await meilisearchClient.index('movies').waitForTask(documentsTask.taskUid)
-  })
-
-  test('one string facet on filter without a query', async () => {
-    const response = await searchClient.search<Movies>([
-      {
-        indexName: 'movies',
-        params: {
-          query: '',
-          facetFilters: ['genres:Adventure'],
-        },
-      },
-    ])
-
-    const hits = response.results[0].hits
-    expect(hits.length).toEqual(1)
-    expect(hits[0].title).toEqual('Star Wars')
-  })
-
-  test('one facet on filter with a query', async () => {
-    const response = await searchClient.search<Movies>([
-      {
-        indexName: 'movies',
-        params: {
-          query: 'four',
-          facetFilters: ['genres:Crime'],
-        },
-      },
-    ])
-
-    const hits = response.results[0].hits
-    expect(hits.length).toEqual(2)
-    expect(hits[0].title).toEqual('Four Rooms')
   })
 
   test('one string facet on filter without a query', async () => {
@@ -173,5 +146,35 @@ describe('Instant Meilisearch Browser test', () => {
 
     const hits = response.results[0].hits
     expect(hits[0].title).toEqual('Ariel')
+  })
+
+  test('numeric filter', async () => {
+    const params = {
+      indexName: 'movies',
+      params: {
+        query: '',
+        numericFilters: 'release_date:593395200 TO 818467200',
+      },
+    }
+
+    const response = await searchClient.search<Movies>([params])
+
+    const hits = response.results[0].hits
+    expect(hits.length).toEqual(3)
+    expect(hits.map((v) => v.release_date).sort()).toEqual([
+      593395200, 750643200, 818467200,
+    ])
+  })
+
+  test('quotation marks and backslashes', () => {
+    const params = {
+      indexName: 'movies',
+      params: {
+        query: '',
+        facetFilters: 'escape_\\_"me":escape \\"me too"',
+      },
+    }
+
+    return expect(searchClient.search<Movies>([params])).resolves.toBeTruthy()
   })
 })
--- a/packages/instant-meilisearch/__tests__/filter.test.ts
+++ b/packages/instant-meilisearch/__tests__/filter.test.ts
@@ -14,8 +14,8 @@ describe('Instant Meilisearch Browser test', () => {
       .updateFilterableAttributes([
         'genres',
         'title',
-        'release_date',
-        'escape_\\_"me"',
+        'numberField',
+        'crazy_\\_"field"',
       ])
     const documentsTask = await meilisearchClient
       .index('movies')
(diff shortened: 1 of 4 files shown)


flevi29:
There are some TODOs in the changes which we might want to address here.

flevi29:
It seems [`String.prototype.replaceAll()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#browser_compatibility) isn't implemented in Node.js 14. Will fix.

However this does bring up the notion that [Node.js 16 is EOL since September 11](https://nodejs.dev/en/about/releases/) (was this date intentional?), so maybe supporting 14 is a little bit of a stretch. But that might be part of another discussion.

packages/instant-meilisearch/__tests__/filter.test.ts
```diff
@@ -50,38 +55,6 @@ describe('Instant Meilisearch Browser test', () => {
...
-          facetFilters: ['genres:Crime'],
-        },
-      },
-    ])
-
-    const hits = response.results[0].hits
-    expect(hits.length).toEqual(2)
-    expect(hits[0].title).toEqual('Four Rooms')
-  })
-
```
brunoocasali, line 84: Why were those tests removed?

packages/instant-meilisearch/__tests__/filter.test.ts
```diff
@@ -174,4 +147,34 @@ describe('Instant Meilisearch Browser test', () => {
     const hits = response.results[0].hits
     expect(hits[0].title).toEqual('Ariel')
   })
+
+  test('numeric filter', async () => {
```
brunoocasali, line 151: I'm curious: why do you have to change the numeric filter handling?

packages/instant-meilisearch/src/adapter/search-request-adapter/filter-adapter.ts
```diff
@@ -1,93 +1,155 @@
...
+function transformFacetFilter(filter: string): string {
+  const escapedFilter = getValueWithEscapedBackslashesAndQuotes(filter)
+  const colonIndex = escapedFilter.indexOf(':')
+  const attribute = escapedFilter.slice(0, colonIndex)
+  const value = escapedFilter.slice(colonIndex + 1)
+  return `"${attribute}"="${value}"`
+}
+
+// Matches first occurrence of an operator
+const numericSplitRegExp = /(?<!(?:[<!>]?=|<|>|:).*)([<!>]?=|<|>|:)/
```
brunoocasali, line 33: Can you provide some tests to cover those different operator cases?

packages/instant-meilisearch/src/adapter/search-request-adapter/filter-adapter.ts
```diff
@@ -1,93 +1,155 @@
...
+ * Wrap the attribute between quotes.
+ * Escape backslash (\\) and quote (") characters.
+ *
+ * @param {string} filter
+ * @returns {string}
  */
-function transformFilter(filter: string): string {
-  return filter.replace(/(.*):(.*)/i, '"$1"="$2"')
+function transformNumericFilter(filter: string): string {
+  // TODO: What if escape facet values is enabled?
```
brunoocasali, line 50: If we merge your PR, those questions will likely never be answered... What can we do about them?

packages/instant-meilisearch/src/adapter/search-request-adapter/filter-adapter.ts
```diff
@@ -1,93 +1,155 @@
...
+    : filters
+        .map((filter) =>
+          (<(value: readonly any[] | any) => value is readonly any[]>(
+            Array.isArray
+          ))(filter)
+            ? filter
+                .map((nestedFilter) => transformCallback(nestedFilter))
+                // TODO: Do these filters have any purpose? Can we actually get empty strings? Should we handle empty strings?
+                //       Maybe handling these should be the responsibility of instantsearch.js and/or the user and their input.
+                //       Instead of these empty filters quietly being swallowed, they might even get us an error message, and
```
brunoocasali, line 91: Can you provide a use test case for that? Because I agree with you, we should not handle that if instantsearch already do it.

packages/instant-meilisearch/src/adapter/search-request-adapter/filter-adapter.ts
```diff
@@ -1,93 +1,155 @@
...
-  numericFilters: Filter,
-  filters: string
+  filters?: string,
+  transformedNumericFilters?: Filter,
+  transformedFacetFilters?: Filter
 ): Filter {
-  const adaptedFilters = filters.trim()
-  const adaptedFacetFilters = filterToArray(facetFilters)
-  const adaptedNumericFilters = filterToArray(numericFilters)
+  // TODO: If we are trimming this, shouldn't we trim the other ones too?
```
brunoocasali, line 133: I guess we should, yes.

brunoocasali (commented):
Hi @flevi29 thanks a lot for your PR!

I'm getting more context for this repo, so I am still cold regarding its quirks and specificities. So, if this PR had only the change of the escaping, it would be amazing, but since there are two different fixes in the same PR, it makes it harder for me to do a proper review. 

Anyway, I added some comments, see you!

Thanks again!

packages/instant-meilisearch/__tests__/filter.test.ts
flevi29: Because they were added twice, it seems.

https://github.com/meilisearch/meilisearch-js-plugins/blob/f390bc953085c63170775233d59d14b37a87d2cf/packages/instant-meilisearch/__tests__/filter.test.ts#L37-L51

https://github.com/meilisearch/meilisearch-js-plugins/blob/f390bc953085c63170775233d59d14b37a87d2cf/packages/instant-meilisearch/__tests__/filter.test.ts#L69-L83

packages/instant-meilisearch/__tests__/filter.test.ts
flevi29: That's because we want to escape the attribute, as it might contain special characters. The old regular expression (`filter.replace(/(.*):(.*)/i, '"$1"="$2"')`) didn't match [numeric comparisons](https://www.algolia.com/doc/api-reference/api-parameters/numericFilters/#numeric-comparisons), so nothing was escaped.

packages/instant-meilisearch/src/adapter/search-request-adapter/filter-adapter.ts
flevi29: I'll look into it more, because I am unsure if trimming is needed at all.

packages/instant-meilisearch/src/adapter/search-request-adapter/filter-adapter.ts
flevi29: What I'm referring to with this, is that under normal circumstances Instantsearch.js shouldn't pass on an empty filter, and if it does, something is clearly wrong. So I propose that we do not check for empty strings, as that might just worsen the issue, because this way we swallow potential errors.

I have to check if by any chance IS.js can return an empty string on purpose, but so far it doesn't seem the case. I'm unsure why the previous maintainer filtered out empty strings, perhaps we can ask him to save us the trouble?

packages/instant-meilisearch/src/adapter/search-request-adapter/filter-adapter.ts
flevi29: Trimming is unnecessary, IS.js should not send trailing spaces, and if it does I believe it's better to leave it as is. Throughout using this patched version, filters was always undefined, and the rest of the filters never had any trailing white space.

packages/instant-meilisearch/src/adapter/search-request-adapter/filter-adapter.ts
brunoocasali: Keep the TODO and you can create a new issue :)

brunoocasali (approved):
I've reproduced the described bug locally and this branch fixes it!

Thanks a lot for your work and patience @flevi29 I've been busy with other priorities. I hope this does not prevent you from keeping contributing!

bors merge

PR #1271, opened 2023-12-17
Re-export `meilisearch` package, and expose created `MeiliSearch` instance

### LINKED ISSUE #1260: Expose both created `MeiliSearch` instance, and the used `meilisearch` package
Currently if one wishes to check, for example, the total number of documents on an index, they need to install the `meilisearch` package as well, and they cannot reuse the already created `MeiliSearch` instance that `@meilisearch/instant-meilisearch` creates. This introduces another potential problem, which is that the version which `@meilisearch/instant-meilisearch` uses and the one that is separately installed is mismatched, resulting in a larger bundle.
This can be avoided by exposing all of `meilisearch` package from `@meilisearch/instant-meilisearch` and also exposing the created instance, sort of like how https://github.com/meilisearch/meilisearch-js-plugins/pull/1258 will expose the search parameters.

flevi29:
# Pull Request

## Related issue
Fixes #1260

## What does this PR do?

- exposed created `MeiliSearch` instance
- re-exported all of `"meilisearch"` package

These improvements make it so that no separate `"meilisearch"` package
has to be installed in order to use its exports directly. This
way a single `MeiliSearch` instance can be re-used, and we can potentially
save on bundle size by avoiding a separate different version installation of
`"meilisearch"`.

```typescript
import { instantMeiliSearch, meilisearch } from '@meilisearch/instant-meilisearch'
// re-exported "meilisearch" ^

const {
  meiliSearchInstance,
  // ^ re-usable MeiliSearch instance
  searchClient
} = instantMeiliSearch(/*...*/)
```


packages/autocomplete-client/scripts/update_version.js
```diff
@@ -1,6 +1,6 @@
 const pkg = require('../package.json')
-const fs = require('fs')
-const path = require('path')
+const fs = require('node:fs')
```
brunoocasali, line 2: Why do you have to make this change? 👀

packages/instant-meilisearch/tsconfig.test.json
```diff
@@ -2,7 +2,6 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "suppressImplicitAnyIndexErrors": true,
-    "resolveJsonModule": true
```
brunoocasali, line 5: 👀 can you explain that 😅 ?

.changeset/sharp-beds-lie.md
```diff
@@ -0,0 +1,23 @@
...
+- re-exported all of `"meilisearch"` package
+
+These improvements make it so that no separate `"meilisearch"` package
+has to be installed in order to use its exports directly. This
+way a single `MeiliSearch` instance can be re-used, and we can potentially
+save on bundle size by avoiding a separate different version installation of
+`"meilisearch"`.
+
+```typescript
+import { instantMeiliSearch, meilisearch } from '@meilisearch/instant-meilisearch'
```
brunoocasali, line 15: Is my suggestion correct right?! 
```suggestion
import { instantMeiliSearch, meiliSearchInstance } from '@meilisearch/instant-meilisearch'
```

packages/autocomplete-client/scripts/update_version.js
flevi29: It's the modern way of importing built-in Node.js modules. The `node:` specifier makes it immediately obvious and non-ambiguous that it's a node built-in module, my IDE complained that `Coding assistance for Node.js is disabled`, so I added this prefix and the warning disappeared.

I'm sorry if I'm making it harder to review these changes, just thought I'd improve a little thing here and there if I'm going to contribute. I can revert these changes if you want to.

Some useful info:
- https://nodejs.org/api/esm.html#node-imports
- https://2ality.com/2021/12/node-protocol-imports.html
- https://docs.deno.com/runtime/manual/node/node_specifiers

packages/instant-meilisearch/tsconfig.test.json
flevi29: It's already set as `true` in the `tsconfig.json` this config file extends:
https://github.com/meilisearch/meilisearch-js-plugins/blob/2bd616385ac42ab55746184cc92b3a83a8e8a2b8/packages/instant-meilisearch/tsconfig.json#L26

- https://www.typescriptlang.org/tsconfig#extends

.changeset/sharp-beds-lie.md
flevi29: The `meilisearch` from `import { instantMeiliSearch, meilisearch } from '@meilisearch/instant-meilisearch'` is the object that contains all of the re-exported elements from `"meilisearch"` package as seen in `index.ts`:
```typescript
// ...
export * as meilisearch from 'meilisearch'
```

The instance is what is created by the main function `instantMeiliSearch()`:

```typescript
instantMeiliSearch(/*...*/).meiliSearchInstance
//                                      ^ this thing, the instance which is also used by the `searchClient`,
//                                          which is then passed to `instantsearch`
```

I could have re-exported it as is (`export * from 'meilisearch'`), but I wanted to avoid any export conflicts, so I made a named re-export `export * as meilisearch from 'meilisearch'`

brunoocasali (approved):
It seems pretty good to me!

brunoocasali:
bors merge

PR #1285, opened 2024-02-05
InstantMeilisearch: Enable hybrid search

Strift:
# Pull Request

Hi 👋 this PR enables using the [hybrid search](https://www.meilisearch.com/docs/learn/experimental/vector_search#vector-search-with-auto-embeddings) experimental API with `@meilisearch/instant-meilisearch`.

## Issue

Solves #1286 

## Proposed API

Allow users to pass a `hybrid` object in the `meiliSearchParams` object when creating an instantMeilisearch client.

```js
instantMeiliSearch(host, apiKey, {
  primaryKey: 'id',
  
  meiliSearchParams: {
    // other meilisearch params
    hybrid: {
      semanticRatio: 0.9,
      embedder: 'default',
    },
  },
})
```

Additionally, the `adaptSearchParams` function will now internally handle this `hybrid` object.

## What does this PR do?
- Add `hybrid` object as an overridable search parameter

## PR checklist
- [x] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [x] Have you read the contributing guidelines?
- [x] Have you made sure that the title is accurate and descriptive of the changes?

### DIFF
--- a/packages/instant-meilisearch/src/adapter/search-request-adapter/search-params-adapter.ts
+++ b/packages/instant-meilisearch/src/adapter/search-request-adapter/search-params-adapter.ts
@@ -232,6 +232,12 @@ export function MeiliParamsCreator(searchContext: SearchContext) {
         meiliSearchParams.attributesToSearchOn = value
       }
     },
+    addHybridSearch() {
+      const value = overrideParams?.hybrid
+      if (value !== undefined) {
+        meiliSearchParams.hybrid = value
+      }
+    },
   }
 }
 
@@ -263,6 +269,7 @@ export function adaptSearchParams(
   meilisearchParams.addMatchingStrategy()
   meilisearchParams.addShowRankingScore()
   meilisearchParams.addAttributesToSearchOn()
+  meilisearchParams.addHybridSearch()
 
   return meilisearchParams.getParams()
 }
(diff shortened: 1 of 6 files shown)


packages/instant-meilisearch/src/types/types.ts
brunoocasali, line 1: Don't you need to make the types for the hybrid search available somewhere?

packages/instant-meilisearch/src/adapter/search-request-adapter/search-params-adapter.ts
```diff
@@ -232,6 +232,12 @@ export function MeiliParamsCreator(searchContext: SearchContext) {
         meiliSearchParams.attributesToSearchOn = value
       }
     },
+    addHybridSearch() {
```
brunoocasali, line 235: Can you add a test case for that?

brunoocasali (changes requested):
Hi @Strift thanks a lot for this PR!

Although, we would need some tests to be able to merge it :)

packages/instant-meilisearch/src/adapter/search-request-adapter/search-params-adapter.ts
Strift: I added a test! It seems to work, but maybe there's a different way to approach this. Let me know!

packages/instant-meilisearch/src/types/types.ts
Strift: I don't think so. 

The `HybridSearch` type is only used internally in `meilisearch` (JS) in the `SearchParams` type.

The `SearchParams` type is exported, so users can still do:

```js
const config: SearchParams[hybrid] = {} // configuration will have the correct type
```

NB: other internal types in `SearchParams`, like `MatchingStrategies` aren't exported either. So I think this is consistent.

PR #1345, opened 2024-12-19
Bump `meilisearch` dependency to 0.47

Strift:
# Pull Request

> **Warning**
> Do not merge before [meilisearch-js](https://github.com/meilisearch/meilisearch-js) releases v0.47

## What does this PR do?
- Bump `meilisearch` dependency to `0.47`


curquiza:
@Strift no need to open the PR in advance, because as you can see [here](https://github.com/meilisearch/meilisearch-js-plugins/pull/1335/files) we need to update yarn.lock too
Bruno will do it the D-day, we don't prepare half of PR in advance 😉 
I close the PR

PR #1347, opened 2024-12-19
Add highlight metadata to autocomplete client

### LINKED ISSUE #1337: Add 'fullyHighlighted', 'matchLevel' and 'matchedWords' to '_highlightResult' in autocomplete client
**Description**
In the hits returned from `getAlgoliaResults`, the `_highlightResult` object includes not just the `value` of the attribute but also three properties that help to interpret the match: `fullyHighlighted`, `matchLevel`, and `matchedWords`. The `getMeilisearchResults` hits, in contrast, only include the `value`. I would like to request the addition of the other three values, if at all possible.

**Basic example**
We are attempting to migrate from Algolia to Meilisearch, but this is a sticking point for us. A real-world example where these properties come into play is one we have frequently: an index of entities that each have a primary name and an array of synonyms, both of which are searchable attributes. In the autocomplete results display, we would like to show just the highlighted name if it is a full match, and if not, we add on the best-matched synonym (showing a fully-highlighted synonym match if one is found, a synonym with matchLevel: full if not, and finally the synonym with the most matched words if no full match level is available).

Here's a real example of what this looks like for us using Algolia's results (showing only the `templates` definition here for brevity):

```js
templates: {
    item({ item, components, html}) {
        if (item._highlightResult.full_name.matchLevel !== 'full' && item._highlightResult.synonyms) {
            var synonym_match = item._highlightResult.synonyms.find(function(synonym) {
                return synonym.matchLevel === 'full' && synonym.fullyHighlighted;
            });
            if (synonym_match === undefined) {
                synonym_match = item._highlightResult.synonyms.find(function(synonym) {
                    return synonym.matchLevel === 'full';
                });
                if (synonym_match === undefined && item._highlightResult.full_name.matchLevel === 'none') {
                    var synonyms = item._highlightResult.synonyms.filter(function (synonym) {
                        return synonym.matchLevel !== 'none';
                    }).sort(function (a, b) {
                        return b.matchedWords.length - a.matchedWords.length;
                    });
                    synonym_match = synonyms[0];
                }
            }
            if (synonym_match !== undefined) {
                synonym_match._highlightResult = { // Workaround to get the highlight component to work (bonus points if you can improve upon Algolia's current default)
                    name: {
                        value: synonym_match.value
                    }
                };
                return html`<span>${components.Highlight({
                    hit: item,
                    attribute: 'full_name',
                })} <span class="text-muted">- synonym match: ${components.Highlight({
                    hit: synonym_match,
                    attribute: 'name',
                })}</span></span>`;
            }
        }

        return html`<span>${components.Highlight({
            hit: item,
            attribute: 'full_name',
        })}</span>`;
    },
},
```

Strift:
# Pull Request

## Related issue
Fixes #1337

## What does this PR do?

This PR enriches the `_highlightResult` object with metadata:
- `fullyHighlighted`:  whether the highlighted text covers the entire field
- `matchLevel`: degree of the match (none, partial, or full)
- `matchedWords`: array of words matched

While adding the necessary tests, I also improved the existing tests for `fetchMeilisearchResults.ts`.

### DIFF
--- a/.changeset/silly-badgers-hug.md
+++ b/.changeset/silly-badgers-hug.md
@@ -1,0 +1,5 @@
+---
+"@meilisearch/autocomplete-client": minor
+---
+
+Add highlight metadata
--- a/packages/autocomplete-client/__tests__/test.utils.ts
+++ b/packages/autocomplete-client/__tests__/test.utils.ts
@@ -17,4 +17,59 @@ const meilisearchClient = new MeiliSearch({
   apiKey: 'masterKey',
 })
 
+export const MOVIES = [
+  {
+    id: 2,
+    title: 'Ariel',
+    overview:
+      "Taisto Kasurinen is a Finnish coal miner whose father has just committed suicide and who is framed for a crime he did not commit. In jail, he starts to dream about leaving the country and starting a new life. He escapes from prison but things don't go as planned...",
+    genres: ['Drama', 'Crime', 'Comedy'],
+    poster: 'https://image.tmdb.org/t/p/w500/ojDg0PGvs6R9xYFodRct2kdI6wC.jpg',
+    release_date: 593395200,
+  },
+  {
+    id: 5,
+    title: 'Four Rooms',
+    overview:
+      "It's Ted the Bellhop's first night on the job...and the hotel's very unusual guests are about to place him in some outrageous predicaments. It seems that this evening's room service is serving up one unbelievable happening after another.",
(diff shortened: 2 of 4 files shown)


Strift:
Tests seem to be failing because I'm testing against Meilisearch 1.12 new highlighting format. Will re-run the CI and look to merge after Meilisearch 1.12 release.

flevi29 (approved):
bors merge

